PR #43, opened 2020-01-07
fix image restoration validation logic

ksnzh:
If trainning image restoration task with DDP, process with `local_rank!= 0` will go to #214 - #260.


xinntao:

PR #55, opened 2020-03-01
support not cached data and lmdb in video_test_dataset

ryul99:
Loading all test images to memory is consuming a lot of memory and this get biggers when GPU distributed training. 
If using not cached data or lmdb in video_test_dataset, memory usage is much smaller than loading all images


xinntao:

PR #71, opened 2020-04-29
add Dockerfile

ryul99:
I made Dockerfile of mmsr repository.
I checked this works properly with EDVR.


xinntao:

Issue #76, opened 2020-05-11
Is there mismatch between GT and selected frames for some special cases in REDS_dataset.py?


SunGaofeng:
Thanks for this excellent repo. I've read some of the codes, and have a question in this line:
In Line 138, 
https://github.com/open-mmlab/mmsr/blob/a73b318f0f07feb6505ef5cb1abf0db33e33807a/codes/data/REDS_dataset.py#L138
If (center_frame_idx > 97) or  (center_frame_idx < 2), center_frame_idx will be generated randomly, center_frame_idx = random.randint(0, 99)

However, the corresponding key for GT is not changed, it will still use the old key to get GT in L156
https://github.com/open-mmlab/mmsr/blob/a73b318f0f07feb6505ef5cb1abf0db33e33807a/codes/data/REDS_dataset.py#L156

Will this lead to mismatch between gt and selected frames when (center_frame_idx > 97) or  (center_frame_idx < 2) ?

For example, if:

    key = '010_00000001'
    name_a, name_b = '010',  '00000001'
    center_frame_idx = 1

Then in Line138, center_frame_idx will be generated randomly, suppose then

    center_frame_idx = 55

A list of neighboring frames will be taken [53, 54, 55, 56, 57]. Then the key of GT for this list should be '010_00000055'. However, in L156, it still takes key='010_00000001'. Will this lead to mismatch?

ryul99:
I'm not collaborator but I think this could lead to mismatch too.

ryul99:
I checked this happen. I tested https://github.com/open-mmlab/mmsr/blob/master/codes/data_scripts/test_dataloader.py with `use_shuffle=False`, `border_mode=False`, `batch_size=1` and setting `shuffle=False` here
https://github.com/open-mmlab/mmsr/blob/a73b318f0f07feb6505ef5cb1abf0db33e33807a/codes/data/__init__.py#L19

xinntao:
Thanks, @SunGaofeng, and @ryul99 
This is a bug!
Also, thank @ryul99  for fixing it~

PR #77, opened 2020-05-12
fix key mismatching with LQ frames

ryul99:
As written in #76, there is a key mismatching bug with LQ frames. This happen when `border_mode=False`.
I think this is a major issue and please review this or check #76 ASAP.


xinntao:
Thanks!

PR #92, opened 2020-07-10
Add readthedocs config files and fix docstring

hejm37:


requirements/readthedocs.txt
```diff
@@ -0,0 +1,3 @@
+mmcv
```
xinntao, line 1: need version?

PR #106, opened 2020-07-17
Add support for concatenating datasets

hejm37:


mmedit/datasets/builder.py
```diff
@@ -32,11 +46,16 @@ def build_dataset(cfg, default_args=None):
     Returns:
```
xinntao, line 46: You may want to update the doc str, also add doc str to _concat_dataset

nbei:
No further comment.

mmedit/datasets/builder.py
hejm37: Though I think for this very function the current docstring is enough, I've added more explanation to the possible behavior of the function.

PR #109, opened 2020-07-18
Fix setup and CI

hellock:


hejm37:
It looks like the github action is not triggered.

PR #136, opened 2020-09-17
[Update] Update version file

nbei:


.gitignore
```diff
@@ -105,7 +105,7 @@ venv.bak/
 .mypy_cache/
 
 # custom
-mmedit/version.py
+# mmedit/version.py
```
hellock, line 108: Just remove this line

mmedit/__init__.py
```diff
@@ -1 +1,30 @@
-from .version import __version__, short_version  # noqa: F401
+import mmcv
+
+from .version import __version__, version_info
+
+MMCV_MIN = '1.0.2'
+MMCV_MAX = '1.2.0'
```
hellock, line 6: 1.2

setup.py
```diff
@@ -54,26 +53,6 @@ def get_hash():
     return sha
```
hejm37, line 53: Looks like the `get_hash` and `get_git_hash` are no longer used?

nbei: I think this func may be called by extra repo. ^_^

hellock (approved):

PR #137, opened 2020-09-18
add collect-env

nbei:


mmedit/utils/collect_env.py
```diff
@@ -0,0 +1,17 @@
+from mmcv.utils import collect_env as collect_base_env
+from mmcv.utils import get_git_hash
+
+import mmedit
+
+
+def collect_env():
```
hellock, line 7: This method is not called before training?

hellock (approved):

PR #138, opened 2020-09-21
add pt1.6 cpu/gpu in CI

nbei:


hellock:
See https://github.com/open-mmlab/mmdetection/blob/master/.github/workflows/build.yml for reference

hellock (approved):

PR #144, opened 2020-09-27
update dockerfile with pt1.6

nbei:


hellock:
Is it tested?

nbei:
> Is it tested?

It's being tested on Tencent Cloud but suffering from low speed when downloading some packages.

hellock (approved):

PR #164, opened 2020-11-26
Update crop.py

wwhio:
modify `Crop` to handle a sequence of video frames


nbei:
Thanks for your contribution. I will have a pass.

PR #184, opened 2021-01-21
a try of docs-refactoring

plyfager:
1. Basiclly follow the given structure.
2. It seems lack some documents, althogh I split some mds into new ones, but it seems not enugh.


docs/modelzoo_statistics.md
```diff
@@ -0,0 +1,161 @@
+
+# Model Zoo
```
congee524, line 2: The file should not be traced.
`modelzoo_statistics.md` will be generated by `stat.py` which is called in `conf.py`.

congee524:
you should modify the file `stat.py` to fix the problem that there are two `Model Zoo` in the content like this 
image

The title in the content is decided by the first-level title in the markdown file. Since the first-level title of `modelzoo_statistics.md` is `Model Zoo`, the title in the content is also `Model Zoo`

nbei:
@plyfager , you may fix the double `module_zoo` problem according to the guidance from @congee524 .

docs/modelzoo_statistics.md
congee524: The file still seems to be tracked.

PR #186, opened 2021-01-26
New algorithm stat

innerlee:
image


innerlee:
Maybe SR should be changed back to restoration? @nbei 

Could you please add links to configs in the READMEs as other repos (mmaction2, mmpose, etc.)?

yaochaorui (approved):

PR #193, opened 2021-02-06
Data prep with extended fg.

yaochaorui:


tools/data/preprocess_comp1k_dataset.py
```diff
@@ -196,6 +197,12 @@ def generate_json(data_root, source_bg_dir, composite, nproc, mode):
         alpha = (
             np.array(Image.open(alpha_full_path).convert('RGB')) /
             255. if composite else None)
+
+        # extend fg
+        image = load_image(fg_full_path, 'RGB')
+        alpha = load_image(alpha_full_path, 'GRAY')
+        F = estimate_foreground_ml(image, alpha, return_background=False)
+        fg = Image.fromarray(np.uint8(F * 255))
```
innerlee, line 205: 1. Is this new fg saved anywhere?
2. Make this functionality optional

yaochaorui: 1. No need to save since the target fg used in the loss should be the gt rather than extended one.
2. Good idea.

innerlee: Okay, notitce that this line overrides L196

yaochaorui: Thanks. If I did not make mistakes, fg is only needed for merging, which can be substituted by extended version. The dump will only store the paths. Thanks for checking.

yaochaorui:
Wait. There should be a better way.

PR #204, opened 2021-02-21
Avoid address in use error.

yaochaorui:


tools/slurm_train.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+export MASTER_PORT=$((12000 + RANDOM % 20000))
```
innerlee, line 2: `RANDOM` -> `$RANDOM`

innerlee (approved):

PR #226, opened 2021-03-16
Add model: multilayer perceptrons.

Yshuo-Li:
Add multilayer perceptrons, which is the `imnet` model employed by LIIF.

### DIFF
--- a/mmedit/models/components/refiners/mlp.py
+++ b/mmedit/models/components/refiners/mlp.py
@@ -1,0 +1,58 @@
+import torch.nn as nn
+from mmcv.runner import load_checkpoint
+
+from mmedit.models.registry import COMPONENTS
+from mmedit.utils import get_root_logger
+
+
+@COMPONENTS.register_module()
+class MLP(nn.Module):
+    """Multilayer perceptrons (MLPs).
+
+    Args:
+        in_dim (int): Input dimension.
+        out_dim (int): Output dimension.
+        hidden_list (list[int]): List of hidden dimensions.
+    """
+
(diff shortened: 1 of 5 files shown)


mmedit/models/components/refiners/mlp.py
```diff
@@ -0,0 +1,58 @@
...
+            pretrained (str, optional): Path for pretrained weights. If given
+                None, pretrained weights will not be loaded. Defaults to None.
+            strict (boo, optional): Whether strictly load the pretrained model.
+                Defaults to True.
+        """
+        if isinstance(pretrained, str):
+            logger = get_root_logger()
+            load_checkpoint(self, pretrained, strict=strict, logger=logger)
+        elif pretrained is None:
+            pass
```
innerlee, line 55: Why it is passed

Yshuo-Li: Because `pretrained is None` means no pretrained model for MLP and there is no special parameter initialization method for MLP.

innerlee: Okay

innerlee (approved):

mmedit/models/components/refiners/mlp.py
innerlee: @nbei what do you think

innerlee:
On the locaiton of the file: is it a kind of `refiner`?

Yshuo-Li:
> On the locaiton of the file: is it a kind of `refiner`?

I am not sure where should the file locate. 
Can you give me a suggest? Thank you.

innerlee (approved):

Issue #205, opened 2021-02-24
Reimplement MaxUnpool2d for onnx


innerlee:
The previous one was reverted.

See https://github.com/open-mmlab/mmediting/pull/196

PR #232, opened 2021-04-01
[feature]: add feature to support dim export onnx model

tangyanf:
this pr is used to support dim to export onnx model


innerlee:
Thanks! This will fix https://github.com/open-mmlab/mmediting/issues/205 by detecting the `is_in_onnx_export` mode, right?

tangyanf:
> Thanks! This will fix #205 by detecting the `is_in_onnx_export` mode, right?

yes,It is only used when exporting the onnx model, and does not affect the training process

innerlee (approved):

PR #243, opened 2021-04-08
fix onnx conversion of maxunpool2d

grimoire:
Hi
This PR fix ONNX conversion of maxunpool2d by adding offset to indices.

### DIFF
--- a/mmedit/models/backbones/encoder_decoders/decoders/plain_decoder.py
+++ b/mmedit/models/backbones/encoder_decoders/decoders/plain_decoder.py
@@ -1,5 +1,3 @@
-import warnings
-
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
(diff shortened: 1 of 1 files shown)


mmedit/models/backbones/encoder_decoders/decoders/plain_decoder.py
```diff
@@ -36,9 +34,62 @@ def forward(ctx, input, indices, kernel_size, stride, padding,
 
     @staticmethod
     def symbolic(g, input, indices, kernel_size, stride, padding, output_size):
-        warnings.warn(
-            'The definitions of indices are different between Pytorch and ONNX'
-            ', so the outputs between Pytorch and ONNX maybe different')
+        # get shape
```
RunningLeon, line 37: We may add the reasons why we do this change with comments here.

hellock (approved):

PR #245, opened 2021-04-08
Add BasicVSR

ckkelvinchan:
BasicVSR: The Search for Essential Components in Video Super-Resolution and Beyond


mmedit/models/restorers/basicvsr.py
```diff
@@ -0,0 +1,179 @@
...
+                 test_cfg=None,
+                 pretrained=None):
+        super(BasicVSR, self).__init__(generator, pixel_loss, train_cfg,
+                                       test_cfg, pretrained)
+
+        # fix pre-trained networks
+        self.fix_iter = train_cfg.get('fix_iter', 0) if train_cfg else 0
+        self.generator.find_unused_parameters = False
+
+        self.step_counter = 0  # count training steps
```
Yshuo-Li, line 44: Does this parameter (`step_counter `) need to be changed when model is resumed?

ckkelvinchan: Yes. It should start with the resume iteration. For instance, if we resume at 10000-th iter, we should have `self.step_counter=10000`. 

Maybe we can pass an argument in `train_cfg`? Or is there any better way to do that?

Yshuo-Li (commented):
Thanks for your contribution! There are two files named `basicvsr.py`, how about rename `mmedit/models/backbones/sr_backbones/basicvsr.py` as `basicvsr_net.py` or others?

ckkelvinchan:
> Thanks for your contribution! There are two files named `basicvsr.py`, how about rename `mmedit/models/backbones/sr_backbones/basicvsr.py` as `basicvsr_net.py` or others?

Sure, I will change it to `basicvsr_net.py`.

mmedit/models/restorers/basicvsr.py
Yshuo-Li: Passing the value from `train_cfg` is a good way.

Yshuo-Li: Passing the value from A is a good idea.

Yshuo-Li: Passing the value from `train_cfg` is a good idea.

mmedit/models/backbones/sr_backbones/basicvsr.py
```diff
@@ -0,0 +1,417 @@
...
+    Args:
+        mid_channels (int): Channel number of the intermediate features.
+            Default: 64.
+        num_blocks (int): Number of residual blocks in each propagation branch.
+            Default: 30.
+        spynet_pretrained (str): Pre-trained model path of SPyNet.
+            Default: None.
+    """
+
+    def __init__(self, mid_channels=64, num_blocks=30, spynet_pretrained=None):
```
Yshuo-Li: Is `spynet_pretrained ` a hard requirement here? 
It seems that `SPyNet` is a sub-model of `BasicVSR` and its parameters cannot be saved separately.

ckkelvinchan: When training BasicVSR, we assume the optical flow network is pretrained. Therefore, we need to separately initialize the weights of SPyNet even if BasicVSR is trained from scratch. Pretrained weights will be provided.

Yshuo-Li: Fine.

mmedit/models/backbones/sr_backbones/basicvsr_net.py
```diff
@@ -0,0 +1,417 @@
...
+            Default: 64.
+        num_blocks (int): Number of residual blocks in each propagation branch.
+            Default: 30.
+        spynet_pretrained (str): Pre-trained model path of SPyNet.
+            Default: None.
+    """
+
+    def __init__(self, mid_channels=64, num_blocks=30, spynet_pretrained=None):
+
+        super(BasicVSRNet, self).__init__()
```
innerlee, line 33: super().__init__()

mmedit/models/backbones/sr_backbones/basicvsr_net.py
```diff
@@ -0,0 +1,417 @@
...
+
+    Args:
+        in_channels (int): Number of input channels of the first conv.
+        out_channels (int): Number of channels of the residual blocks.
+            Default: 64.
+        num_blocks (int): Number of residual blocks. Default: 30.
+    """
+
+    def __init__(self, in_channels, out_channels=64, num_blocks=30):
+        super(ResidualBlocksWithInputConv, self).__init__()
```
innerlee, line 196:
```python
super().__init__()
```

mmedit/models/backbones/sr_backbones/basicvsr_net.py
```diff
@@ -0,0 +1,417 @@
...
+
+    Paper:
+        Optical Flow Estimation using a Spatial Pyramid Network, CVPR, 2017
+
+    Args:
+        pretrained (str): path for pre-trained SPyNet. Default: None.
+    """
+
+    def __init__(self, pretrained):
+        super(SPyNet, self).__init__()
```
innerlee, line 239: init

mmedit/models/restorers/basicvsr.py
```diff
@@ -0,0 +1,182 @@
...
+        pretrained (str): Path for pretrained model. Default: None.
+    """
+
+    def __init__(self,
+                 generator,
+                 pixel_loss,
+                 train_cfg=None,
+                 test_cfg=None,
+                 pretrained=None):
+        super(BasicVSR, self).__init__(generator, pixel_loss, train_cfg,
```
innerlee, line 37: super

mmedit/models/restorers/basicvsr.py
```diff
@@ -0,0 +1,182 @@
...
+                                       test_cfg, pretrained)
+
+        # fix pre-trained networks
+        self.fix_iter = train_cfg.get('fix_iter', 0) if train_cfg else 0
+        self.generator.find_unused_parameters = False
+
+        # count training steps
+        self.step_counter = train_cfg.get('step_counter',
```
innerlee, line 45: Proposal:

Add a  `TAIL` component to the module, just like backbones and heads, and save step counters there, so that they could be automatically saved and restored.

mmedit/models/restorers/basicvsr.py
```diff
@@ -0,0 +1,182 @@
...
+
+        Returns:
+            dict: Output results.
+        """
+        with torch.no_grad():
+            output = self.generator(lq)
+
+        # Note: For Vimeo-90K, we use mirror extension. Hence, the output
+        # sequence has 14 frames. Only center frame is kept for Vimeo-90K.
+        if output.size(1) == 14:
```
innerlee, line 141: I don't think the hard-coded 14 is good

innerlee:
Looks good generally. Some minor comments.

innerlee (approved):

PR #250, opened 2021-04-12
Support Y-channel PSNR and SSIM

ckkelvinchan:
Many existing SR methods compute PSNR and SSIM on Y-channel. However, MMEditing currently does not support this operation. This PR adds this operation by adding an argument `color_space` in `psnr` and `ssim`.

### DIFF
--- a/mmedit/core/evaluation/metrics.py
+++ b/mmedit/core/evaluation/metrics.py
@@ -165,7 +165,7 @@ def reorder_image(img, input_order='HWC'):
     return img
 
 
-def psnr(img1, img2, crop_border=0, input_order='HWC'):
+def psnr(img1, img2, crop_border=0, input_order='HWC', color_space=None):
     """Calculate PSNR (Peak Signal-to-Noise Ratio).
 
     Ref: https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio
@@ -177,6 +177,8 @@ def psnr(img1, img2, crop_border=0, input_order='HWC'):
             pixels are not involved in the PSNR calculation. Default: 0.
         input_order (str): Whether the input order is 'HWC' or 'CHW'.
             Default: 'HWC'.
+        color_space (str): The color space in which PSNR is computed. If None,
+            the images are not altered. Default: None.
 
     Returns:
         float: psnr result.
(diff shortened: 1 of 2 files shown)


mmedit/core/evaluation/metrics.py
```diff
@@ -269,6 +281,16 @@ def ssim(img1, img2, crop_border=0, input_order='HWC'):
     img1 = reorder_image(img1, input_order=input_order)
     img2 = reorder_image(img2, input_order=input_order)
 
+    if isinstance(color_space, str) and color_space.lower() == 'Y':
```
innerlee, line 284: Whose lower equals "Y"?

mmedit/core/evaluation/metrics.py
```diff
@@ -177,6 +177,8 @@ def psnr(img1, img2, crop_border=0, input_order='HWC'):
             pixels are not involved in the PSNR calculation. Default: 0.
         input_order (str): Whether the input order is 'HWC' or 'CHW'.
             Default: 'HWC'.
+        color_space (str): The color space in which PSNR is computed. If None,
+            the images are not altered. Default: None.
```
innerlee, line 181: Give the full supported options here

innerlee: Another note: when computing for "Y", img1 and img2 are assumed in BGR order

mmedit/core/evaluation/metrics.py
```diff
@@ -165,7 +165,7 @@ def reorder_image(img, input_order='HWC'):
     return img
 
 
-def psnr(img1, img2, crop_border=0, input_order='HWC'):
+def psnr(img1, img2, crop_border=0, input_order='HWC', color_space=None):
```
innerlee, line 168: The name "color_space" needs a better design.
"Y" is not a color (standard) space

ckkelvinchan: How about `convert_to`? This is what `niqe` used.

innerlee: convert_to is not good either, think harder

innerlee: Okay, convert_to is ok

ckkelvinchan: Since we either keep the input as it is or turn it to y-channel. How about we use a boolean variable named `to_y_channel`? Not sure if there is a need to support other color spaces in the future.

innerlee: Yes this should be extensible to other channels. Sometimes we want to compute on Hue or Lightness.
Either we do not support color conversion and extraction in this function and users need to convert them before call this function, or we leave a generic interface (no need to actually implement) so that any possible conversion is possible to implement.

innerlee (approved):

PR #253, opened 2021-04-13
Add MirrorExtend for training BasicVSR and IconVSR

ckkelvinchan:
BasicVSR and IconVSR extend the 7-frame Vimeo-90K sequences to 14 frames by mirroring them. This PR adds support to this operation by adding `MirrorExtend`.

### DIFF
--- a/mmedit/datasets/pipelines/augmentation.py
+++ b/mmedit/datasets/pipelines/augmentation.py
@@ -898,7 +898,7 @@ class TemporalReverse:
 
 
 @PIPELINES.register_module()
-class MirrorSequenceExtend:
+class MirrorSequence:
     """Extend short sequences (e.g. Vimeo-90K) by mirroring the sequences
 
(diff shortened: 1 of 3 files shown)


mmedit/datasets/pipelines/augmentation.py
```diff
@@ -895,3 +895,42 @@ def __repr__(self):
         repr_str = self.__class__.__name__
         repr_str += f'(keys={self.keys}, reverse_ratio={self.reverse_ratio})'
         return repr_str
+
+
+@PIPELINES.register_module()
+class MirrorExtend:
```
innerlee, line 901: How about `MirrorSequenceExtend`, so that to rule out the wrong guess of spatial extention.

ckkelvinchan: This one seems more succinct. Will change~Thanks~

innerlee (approved):

PR #265, opened 2021-04-21
[Feature] Update pytorch2onnx

grimoire:
Hi,
This PR adds restorer model support to pytorch2onnx.
Dynamic shape, dynamic batch and visualize support are also added too.

### DIFF
--- a/tools/pytorch2onnx.py
+++ b/tools/pytorch2onnx.py
@@ -55,7 +55,7 @@ def pytorch2onnx(model,
                 3: 'width'
             },
             'output': {
-                1: 'batch',
+                0: 'batch',
                 2: 'height',
                 3: 'width'
             }
--- a/tools/pytorch2onnx.py
+++ b/tools/pytorch2onnx.py
@@ -1,5 +1,6 @@
 import argparse
 
+import matplotlib.pyplot as plt
 import mmcv
 import numpy as np
 import onnx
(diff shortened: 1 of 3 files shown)


tools/pytorch2onnx.py
```diff
@@ -12,12 +13,28 @@
 from mmedit.models import build_model
 
 
+def show_result_pyplot(img, title='', block=True):
```
innerlee, line 16: So this function simply shows an image? Can you rewrite it using cv2.imshow() or imwrite?

innerlee (approved):

tools/pytorch2onnx.py
```diff
@@ -32,85 +35,144 @@ def pytorch2onnx(model,
...
     model.forward = model.forward_dummy
     # pytorch has some bug in pytorch1.3, we have to fix it
     # by replacing these existing op
     register_extra_symbolics(opset_version)
+    dynamic_axes = None
+    if dynamic_export:
+        dynamic_axes = {
+            'input': {
+                0: 'batch',
+                2: 'height',
```
RunningLeon, line 54: Do we support dynamic shape feature for ESRGAN and SRCNN in this tool?

tools/pytorch2onnx.py
```diff
@@ -32,85 +35,144 @@ def pytorch2onnx(model,
...
+                3: 'width'
+            },
+            'output': {
+                1: 'batch',
```
RunningLeon, line 58: Index of batch for output tensor is `1` or `0` ?

RunningLeon (approved):
LGTM

PR #271, opened 2021-04-23
Fix usage of channel_order in loading.py

ckkelvinchan:
In the previous version, `LoadImageFromFileList`  and  `LoadPairedImageFromFile` did not use `channel_order` in `mmcv.imfrombytes`. Therefore, the input is still in `bgr` order even if `channel_order='rgb` is specified in the configuration file. 

This PR adds the argument to the function. The default value is `bgr`, and hence it does not affect the existing codes.


ckkelvinchan:
@innerlee Please note that `RandomLoadResizeBg` also did not use `channel_order`, but I am not sure whether this is necessary. Therefore I leave it unchanged.

innerlee:
`RandomLoadResizeBg` is safe, since it can pass any kwargs (including channel_order of course) to the file client.

innerlee:
@Yshuo-Li This bug will affect the checkpoint of LIIF, because the channel_order in the config is not passed to image decoder. So the model accepts BGR instead of RGB

ckkelvinchan:
> ~`RandomLoadResizeBg` is safe, since it can pass any kwargs (including channel_order of course) to the file client.~
> 
> edit: sorry no. Its imfrombytes. Yes please add the option to RandomLoadResizeBg, it is safe since there is no config actually uses it yet.

Okay, will do

Yshuo-Li:
> @Yshuo-Li This bug ~will affect the checkpoint of LIIF~, because the channel_order in the config is not passed to image decoder. So the model accepts BGR instead of RGB

LIIF uses `LoadImageFromFile` instead of `LoadImageFromFileList`
In `LoadImageFromFile` line 73:
`img = mmcv.imfrombytes(
                img_bytes, flag=self.flag,
                channel_order=self.channel_order)`

innerlee:
@ckkelvinchan since this change will affect the checkpoint, do you want to re-train one?

ckkelvinchan:
> @ckkelvinchan since this change will affect the checkpoint, do you want to re-train one?

I am reproducing GLEAN, so I may need to do it later. I suggest changing the configuration files for BasicVSR and IconVSR back to `bgr` for the moment.

innerlee (approved):

wwhio:
> I suggest changing the configuration files for BasicVSR and IconVSR back to bgr for the moment.

Hello @ckkelvinchan , thank you for your awesome work. Will this change affect the optical flow estimated by SPyNet ? According to [this code](https://github.com/sniklaus/pytorch-spynet/blob/3f1a8abda37f20fd1b33cf1d9122dc8be4189686/run.py#L169), SPyNet takes an image with `bgr` channel order.

ckkelvinchan:
The SPyNet we used takes RGB as inputs. This is slightly different from the one you mentioned.

wwhio:
OK. Thank you.

PR #274, opened 2021-04-24
Update weights and re-evaluate models

ckkelvinchan:
When `channel_order` is not used correctly, BasicVSR and IconVSR output BGR image hence the evaluation result is not correct (since the function `bgr2gray` is used when evaluating in Y-channel). This PR re-computes the results in the datasets and updates the links for the pre-trained weights.


configs/restorers/iconvsr/iconvsr_vimeo90k_bd.py
```diff
@@ -16,7 +16,7 @@
     pixel_loss=dict(type='CharbonnierLoss', loss_weight=1.0, reduction='mean'))
 # model training and testing settings
 train_cfg = dict(fix_iter=5000)
-test_cfg = dict(metrics=['PSNR', 'SSIM'], crop_border=0, convert_to='y')
```
innerlee, line 19: Is removing 'SSIM' intended?

ckkelvinchan: Yes, as it would slow down the training. Therefore I put only PSNR here.

ckkelvinchan:
Invalid link, will update

configs/restorers/iconvsr/iconvsr_vimeo90k_bd.py
innerlee: Users may remove 'SSIM', or together with 'PSNR' themselves if they want to speed up training. It's better to leave it here since our result page contains both SSIM and PSNR.

innerlee (approved):

PR #284, opened 2021-04-29
Update README.md

Yshuo-Li:


innerlee:
CN README needs update as well

Yshuo-Li:
> CN README needs update as well
Will update CN README after finish EN README updating.

innerlee (approved):

PR #301, opened 2021-05-12
[Feature] ImageNormalize

Yshuo-Li:
Mean Shift Layer


mmedit/models/components/layers/mean_shift.py
```diff
@@ -0,0 +1,27 @@
+import torch
+import torch.nn as nn
+
+from mmedit.models.registry import COMPONENTS
+
+
+@COMPONENTS.register_module()
+class MeanShift(nn.Conv2d):
+    """Mean shift.
```
innerlee, line 9: This is too sloppy

mmedit/models/components/layers/mean_shift.py
```diff
@@ -0,0 +1,27 @@
...
+
+    Args:
+        pixel_range (float): Pixel range of image.
+        rgb_mean (Tuple[float]): Image mean in RGB orders.
+        rgb_std (Tuple[float]): Image std in RGB orders.
+        sign (int): Sign of bias. Default -1.
+    """
+
+    def __init__(self, pixel_range, rgb_mean, rgb_std, sign=-1):
+        super(MeanShift, self).__init__(3, 3, kernel_size=1)
```
innerlee, line 19: `super().__init__(...)`

mmedit/models/components/layers/mean_shift.py
```diff
@@ -0,0 +1,27 @@
...
+        std = torch.Tensor(rgb_std)
```
innerlee, line 20: What's the device of this tensor

mmedit/models/components/layers/mean_shift.py
```diff
@@ -0,0 +1,27 @@
+import torch
+import torch.nn as nn
+
+from mmedit.models.registry import COMPONENTS
+
+
+@COMPONENTS.register_module()
+class MeanShift(nn.Conv2d):
```
innerlee, line 8: MeanShift is a classical alogrithm, so I do not like this name. Do you have alternative choices? Like `FeatureShift` or so?

mmedit/models/components/layers/mean_shift.py
Yshuo-Li: CPU.
It is not important, only used in init.

innerlee: No this should be taken care, becuase it might be run on gpu

Yshuo-Li: How about `del std` after __init__?

mmedit/models/components/layers/feature_shift.py
```diff
@@ -0,0 +1,31 @@
+import torch
+import torch.nn as nn
+
+from mmedit.models.registry import COMPONENTS
+
+
+@COMPONENTS.register_module()
+class FeatureShift(nn.Conv2d):
```
innerlee, line 8: As discussed in private, it is a feature normalization. So "FeatureNormalize" maybe more suitable.

innerlee: ImageNormalize

mmedit/models/components/layers/mean_shift.py
innerlee: My understanding was wrong. No need to change :)

innerlee (approved):

Issue #306, opened 2021-05-14
test flops and parameters 


luoyinTT:
when i try to use the tools/get_flops.py to compute flops and parameters follow the guide , i met

```bash
(open-mmlab) yinl@xcdl190206:/scratch/workspace/yinl/mmediting-master$ python tools/get_flops.py configs/restorers/basicvsr/basicvsr_vimeo90k_bi.py --shape 180 120 
2021-05-14 10:39:09,790 - mmedit - INFO - Use load_from_local loader
lrs.size(): torch.Size([1, 3, 180, 120])
Traceback (most recent call last):
  File "tools/get_flops.py", line 57, in <module>
    main()
  File "tools/get_flops.py", line 47, in main
    flops, params = get_model_complexity_info(model, input_shape)
  File "/scratch/workspace/yinl/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmcv/cnn/utils/flops_counter.py", line 104, in get_model_complexity_info
    _ = flops_model(batch)
  File "/scratch/workspace/yinl/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/torch/nn/modules/module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/var/lib/docker/scratch/workspace/yinl/mmediting-master/mmedit/models/restorers/basic_restorer.py", line 174, in forward_dummy
    out = self.generator(img)
  File "/scratch/workspace/yinl/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/torch/nn/modules/module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/var/lib/docker/scratch/workspace/yinl/mmediting-master/mmedit/models/backbones/sr_backbones/basicvsr_net.py", line 117, in forward
    n, t, c, h, w = lrs.size()
ValueError: not enough values to unpack (expected 5, got 4)
```

how to solve it?

ckkelvinchan:
Since BasicVSR takes a video (instead of an image) as input, the existing get_flops.py may be incompatible. I will look into the problem and get back to you soon. Thanks for reporting.

ckkelvinchan:
Fixed in #309. Please reopen if there are any further problems.

valleballe:
I have a similar issue running restoration_video_demo.py 

I followed the installation instructions:
```
pip3 install openmim
mim install mmcv-full
git clone https://github.com/open-mmlab/mmediting.git
[...]

PR #309, opened 2021-05-14
[Feature] Support FLOPs calculation of video SR models

ckkelvinchan:
As discussed in #306,  Existing `get_flops.py` does not accept video inputs with size `(t, c, h, w)`. Therefore, the function is not compatible with video SR models such as EDVR, BasicVSR.

This PR adds the support of FLOPs computation of video SR models, by accepting `--shape t c h w`. For example, for BasicVSR, we can compute the FLOPs by using the following command.

```python tools/get_flops.py configs/restorers/basicvsr/basicvsr_vimeo90k_bi.py --shape 5 3 180 320```

### DIFF
--- a/tools/get_flops.py
+++ b/tools/get_flops.py
@@ -49,8 +49,10 @@ def main():
     split_line = '=' * 30
     print(f'{split_line}\nInput shape: {input_shape}\n'
           f'Flops: {flops}\nParams: {params}\n{split_line}')
-    print('!!!If you are testing a recurrent network, please divide the '
-          'FLOPs by the number of frames.')
+    if len(input_shape) == 4:
+        print('!!!If your network computes N frames in one forward pass, you '
+              'may want to divide the FLOPs by N to get the average FLOPs '
+              'for each frame.')
     print('!!!Please be cautious if you use the results in papers. '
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tools/get_flops.py [hunk 1]
```diff
@@ -45,9 +45,12 @@ def main():
             f'with {model.__class__.__name__}')
 
     flops, params = get_model_complexity_info(model, input_shape)
+
     split_line = '=' * 30
     print(f'{split_line}\nInput shape: {input_shape}\n'
           f'Flops: {flops}\nParams: {params}\n{split_line}')
+    print('!!!If you are testing a recurrent network, please divide the '
```
innerlee, line 52: Why it should be divided?

ckkelvinchan: Since a recurrent network (e.g. BasicVSR) takes N frames as inputs and outputs N frames, therefore the FLOPs for each frame should be the output FLOPs divided by N.

innerlee: Okay, print this info only when the input is N frames, and change "recurrent network" to some other name. N frame to N frame does not have to be recurrent

innerlee (approved):

PR #311, opened 2021-05-16
[Feature] Add Structural Feature Encoder

Yshuo-Li:
Backbone of Texture Transformer Network for Image Super-Resolution.

### DIFF
--- a/mmedit/models/backbones/sr_backbones/ttsr_net.py
+++ b/mmedit/models/backbones/sr_backbones/ttsr_net.py
@@ -1,31 +1,14 @@
+from functools import partial
+
 import torch.nn as nn
 import torch.nn.functional as F
 from mmcv.cnn import build_conv_layer
 
 from mmedit.models.common import ResidualBlockNoBN, make_layer
 
-
-def norm_conv_layer(in_channels, out_channels, stride=1):
-    """Norm conv layer with kernel_size=3.
-
-    Args:
-        in_channels (int): Number of channels in the input image
-        out_channels (int): Number of channels produced by the convolution
-        stride (int or tuple, optional): Stride of the convolution. Default: 1
-
-    results:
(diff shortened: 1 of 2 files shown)


tests/test_ttsr.py
```diff
@@ -0,0 +1,14 @@
+import torch
+
+from mmedit.models.backbones.sr_backbones.ttsr_net import SFE
+
+
+def test_ttsr():
```
innerlee, line 6: ttsr -> sfe

mmedit/models/backbones/sr_backbones/ttsr_net.py
```diff
@@ -0,0 +1,71 @@
+import torch.nn as nn
+import torch.nn.functional as F
+
+from mmedit.models.common import ResidualBlockNoBN, make_layer
+
+
+def norm_conv_layer(in_channels, out_channels, stride=1):
```
innerlee, line 7: mmcv has both ConvModule and build_conv_layer, see examples in https://github.com/open-mmlab/mmclassification/blob/master/mmcls/models/backbones/resnet.py

mmedit/models/backbones/sr_backbones/ttsr_net.py
```diff
@@ -0,0 +1,71 @@
...
+        """Forward function.
+
+        Args:
+            x (Tensor): Input tensor with shape (n, c, h, w).
+
+        Returns:
+            Tensor: Forward results.
+        """
+
+        x = F.relu(self.conv_head(x))
```
innerlee, line 66: The head/body/tail thing is in the reverse order with our convention. In ours, the head is near to the output.

innerlee: I suggest just use the good old conv1 conv2 conv3, etc.

innerlee (approved):

PR #312, opened 2021-05-17
[Feature] Add Cross-Scale Feature Integration

Yshuo-Li:
Part of Texture Transformer Network for Image Super-Resolution.

### DIFF
--- a/mmedit/models/backbones/sr_backbones/ttsr_net.py
+++ b/mmedit/models/backbones/sr_backbones/ttsr_net.py
@@ -59,10 +59,14 @@ class SFE(nn.Module):
 
 
 class CSFI2(nn.Module):
-    """Cross-Scale Feature Integration, include 2 scales.
+    """Cross-Scale Feature Integration between 1x and 2x features.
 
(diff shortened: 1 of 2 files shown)


mmedit/models/backbones/sr_backbones/ttsr_net.py
```diff
@@ -9,6 +10,8 @@
 # Use partial to specify some default arguments
 _norm_conv_layer = partial(
     build_conv_layer, dict(type='Conv2d'), kernel_size=3, padding=1)
+_bottleneck_layer = partial(
```
innerlee, line 13: I'm not sure if **one** conv could be called a "bottleneck"

innerlee: Also double check if normalization in included

Yshuo-Li: Renamed to `_conv1x1_layer`

innerlee (approved):

PR #324, opened 2021-05-24
Add error message in restoration_demo.py

ckkelvinchan:
## Motivation
There are currently two restoration demos. `restoration_demo.py` is for image restoration, and `restoration_video_demo.py` is for video restoration. These two may be confusing. 

## Modification
In this PR `restoration_demo.py` raises an error if `img_path` does not end with `.png` or `.jpg`. In the error message, we suggest the use of `restoration_video_demo.py` for video restoration.

### DIFF
--- a/demo/restoration_demo.py
+++ b/demo/restoration_demo.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 
 import mmcv
 import torch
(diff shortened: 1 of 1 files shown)


demo/restoration_demo.py
```diff
@@ -23,6 +23,12 @@ def parse_args():
 def main():
     args = parse_args()
 
+    if not args.img_path.endswith(('.jpg', '.png')):
```
innerlee, line 26: Use os.path.isfile

innerlee (approved):

Issue #323, opened 2021-05-24
Error while running BasicVSR demo 


Kadakol:
Hi,

Thank you so much for your work on BasicVSR!

Installation was done as per the [guide](https://github.com/open-mmlab/mmediting/blob/master/docs/install.md)
I was trying to run the demo as per this [document](https://mmediting.readthedocs.io/en/latest/demo.html)

Following is the command I used:

`python demo/restoration_demo.py configs/restorers/basicvsr/basicvsr_reds4.py models/basicvsr_reds4_20120409-0e599677.pth <PATH_TO_DATA>/calendar/ ./basic_vsr/vid4/calendar`

Unfortunately, I get the following error:

```
2021-05-24 15:11:05,378 - mmedit - INFO - Use load_from_http loader
Use load_from_local loader
The model and loaded state dict do not match exactly

missing keys in source state_dict: step_counter

Traceback (most recent call last):
  File "demo/restoration_demo.py", line 38, in <module>
    main()
  File "demo/restoration_demo.py", line 29, in main
    output = restoration_inference(model, args.img_path)
  File "<PATH>/mmediting/mmedit/apis/restoration_inference.py", line 35, in restoration_inference
    data = test_pipeline(data)
  File "<PATH>/mmediting/mmedit/datasets/pipelines/compose.py", line 41, in __call__
    data = t(data)
  File "<PATH>/mmediting/mmedit/datasets/pipelines/augmentation.py", line 946, in __call__
    clip_name = results['key']
KeyError: 'key'
Segmentation fault (core dumped)
```

Please suggest how to fix this issue. Also do let me know if any information is required from my side.

Thanks!

ckkelvinchan:
You should use `restoration_video_demo.py` instead of `restoration_demo.py`.

innerlee:
Add some friendly warnings in `restoration_demo.py` if multiple frames are detected? @ckkelvinchan

ckkelvinchan:
> Add some friendly warnings in `restoration_demo.py` if multiple frames are detected? @ckkelvinchan

That is a good idea.

Kadakol:
Hi @ckkelvinchan ,

Thanks for such a quick response! That was very silly of me. I misread the instructions! :(

I tried the following command:
[...]

PR #325, opened 2021-05-24
[Feature] Support different filename templates in GenerateSegmentIndices

ckkelvinchan:
## Motivation
As mentioned in issue #323, there will be errors if the file names are not formated as `{v:08d}.png`. This may cause problems if one use custom filename format.

## Modification
In this PR, we modify `GenerateSegmentIndices` so that one can use different filename formats. But note that the frame indices must start from 0.

### DIFF
--- a/mmedit/apis/restoration_video_inference.py
+++ b/mmedit/apis/restoration_video_inference.py
@@ -27,6 +27,7 @@ def restoration_video_inference(model, img_dir, window_size, filename_tmpl):
         window_size (int): The window size used in sliding-window framework.
             This value should be set according to the settings of the network.
             A value smaller than 0 means using recurrent framework.
+
     Returns:
         Tensor: The predicted restoration result.
     """
--- a/mmedit/apis/restoration_video_inference.py
+++ b/mmedit/apis/restoration_video_inference.py
@@ -20,6 +20,7 @@ def pad_sequence(data, window_size):
 
 def restoration_video_inference(model, img_dir, window_size, filename_tmpl):
     """Inference image with the model.
+
     Args:
         model (nn.Module): The loaded model.
         img_dir (str): Directory of the input video.
@@ -29,6 +30,7 @@ def restoration_video_inference(model, img_dir, window_size, filename_tmpl):
     Returns:
         Tensor: The predicted restoration result.
     """
+
     device = next(model.parameters()).device  # model device
 
     # pipeline
--- a/mmedit/apis/restoration_video_inference.py
+++ b/mmedit/apis/restoration_video_inference.py
@@ -18,16 +18,14 @@ def pad_sequence(data, window_size):
     return data
 
 
-def restoration_video_inference(model, img_dir, window_size):
+def restoration_video_inference(model, img_dir, window_size, filename_tmpl):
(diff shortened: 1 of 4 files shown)


mmedit/apis/restoration_video_inference.py
```diff
@@ -18,24 +18,25 @@ def pad_sequence(data, window_size):
     return data
 
 
-def restoration_video_inference(model, img_dir, window_size):
+def restoration_video_inference(model, img_dir, window_size, filename_tmpl):
```
innerlee, line 21: Have you changed all references to this function?

innerlee: Okay it seems so

mmedit/apis/restoration_video_inference.py
```diff
@@ -18,24 +18,25 @@ def pad_sequence(data, window_size):
...
     """Inference image with the model.
-
```
innerlee, line 23: Do not remove the space

mmedit/apis/restoration_video_inference.py
ckkelvinchan: Yes, only `restoration_video_demo.py` uses this fuction.

mmedit/apis/restoration_video_inference.py
```diff
@@ -27,15 +27,18 @@ def restoration_video_inference(model, img_dir, window_size):
         window_size (int): The window size used in sliding-window framework.
             This value should be set according to the settings of the network.
             A value smaller than 0 means using recurrent framework.
-
```
innerlee, line 30: and here, of course

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -964,11 +966,13 @@ def __call__(self, results):
         lq_path_root = results['lq_path']
         gt_path_root = results['gt_path']
         lq_path = [
-            osp.join(lq_path_root, clip_name, f'{v:08d}.png')
+            osp.join(lq_path_root, clip_name,
+                     f'{self.filename_tmpl.format(v)}.png')
```
innerlee, line 970: Question: should `.png` be part of the template?

ckkelvinchan: We usually use `.png`, but I think that could work if someone wants to use `jpg` as inputs. I will modify that.

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -964,12 +966,12 @@ def __call__(self, results):
         lq_path_root = results['lq_path']
         gt_path_root = results['gt_path']
         lq_path = [
-            osp.join(lq_path_root, clip_name, f'{v:08d}.png')
-            for v in neighbor_list
+            osp.join(lq_path_root, clip_name,
+                     f'{self.filename_tmpl.format(v)}') for v in neighbor_list
```
innerlee, line 970: In this case, the f-str is redundant

ckkelvinchan: Oh you are right. I missed that.

innerlee (approved):

PR #336, opened 2021-05-28
[Feature] Generate Heatmap

Yshuo-Li:


mmedit/datasets/pipelines/generate_assistant.py
```diff
@@ -5,6 +5,78 @@
...
+            landmark (Tuple[float]): Location of a landmark.
+
+        results:
+            heatmap (np.ndarray): A heatmap of landmark.
+        """
+        w, h = self.target_size
+
+        x_range = np.arange(start=0, stop=w, dtype=int)
+        y_range = np.arange(start=0, stop=h, dtype=int)
+        xx, yy = np.meshgrid(x_range, y_range)
```
ckkelvinchan, line 67: Not sure if `xx` and `yy` are good names. How about `grid_x` and `grid_y`?

mmedit/datasets/pipelines/generate_assistant.py
```diff
@@ -5,6 +5,78 @@
...
+        d2 = (xx - keypoint[0])**2 + (yy - keypoint[1])**2
```
ckkelvinchan, line 68: `d2` -> `dist`?

innerlee: Actually here d2 is better than dist, since it is distance squared.

innerlee:
Unrelated comment: we may find a way (maybe in next month) to reduce code duplication between mmedit and mmpose

innerlee (approved):

mmedit/datasets/pipelines/generate_assistant.py
```diff
@@ -5,6 +5,78 @@
...
+    def __init__(self, keypoint, ori_size, target_size, sigma=1.0):
+        if isinstance(ori_size, int):
+            ori_size = (ori_size, ori_size)
+        else:
+            ori_size = ori_size[:2]
+        if isinstance(target_size, int):
+            target_size = (target_size, target_size)
+        else:
+            target_size = target_size[:2]
+        self.rate = (target_size[0] / ori_size[0],
```
innerlee, line 28: What does rate mean here

mmedit/datasets/pipelines/generate_assistant.py
```diff
@@ -5,6 +5,78 @@
...
+        grid_x, grid_y = np.meshgrid(x_range, y_range)
+        dist2 = (grid_x - keypoint[0])**2 + (grid_y - keypoint[1])**2
+        exponent = dist2 / 2.0 / self.sigma / self.sigma
+        heatmap = np.exp(-exponent)
+        return heatmap
+
+    def __repr__(self):
+        return (f'{self.__class__.__name__}, '
+                f'keypoint={self.keypoint}, '
+                f'rate={self.rate}, target_size={self.target_size}, '
```
innerlee, line 76: Do not need to show rate here, since it is not a configurable parameter

mmedit/datasets/pipelines/generate_assistant.py
Yshuo-Li: Resize heatmap

innerlee: Okay should be ratio. Choose `target_ori_ratio`/`size_ratio`/...

Issue #335, opened 2021-05-28
restoration_video_demo.py --filename_tmpl 参数问题


JoshonSmith:
python demo/restoration_video_demo.py ./configs/restorers/basicvsr/basicvsr_vimeo90k_bi.py ./work_dirs/basicvsr_vimeo90k_bi_20210409-d2d8f760.pth ./data/vimeo_super_resolution_test/low_resolution/00001/0268/ ./output/basicvsr/vimeo --device=1 --filename_tmpl='im{:d}.png'

vimeo90k 数据集 的文件格式：im1.png ... im7.png
没有 im0.png
而--filename_tmpl参数该如何指定这种情况？

JoshonSmith:
the frame indices must start from 0 ?

ckkelvinchan:
> the frame indices must start from 0 ?

Yes, since in `GenerateSegmentIndices`, we generate the indices based on the assumption that the first frame has an index 0.

Please rename the files for the moment, we will consider adding an optional argument `start_index` in `GenerateSegmentIndices`.

JoshonSmith:
ok
thanks!

PR #337, opened 2021-05-28
[Feature] Support arbitrary naming in GenerateSegmentIndices

ckkelvinchan:
## Motivation
As mentioned in #335, the existing video_restoration_demo does not support sequences whose start frame index is not 0. This may cause problem in custom video sequences.

## Modification
This PR changes the behavior of the function `GenerateSegmentIndices` so that the image names can now be arbitrary (the latter frames must have larger indices).
1. Instead of assuming the start frame index is 0, `mmcv.utils.scandir` is used to retrieve the filenames.
2. The use of `filename_tmpl` is removed.

In addition, the `video_restoration_inference` is also modified.

### DIFF
--- a/mmedit/datasets/pipelines/augmentation.py
+++ b/mmedit/datasets/pipelines/augmentation.py
@@ -927,12 +927,10 @@ class GenerateSegmentIndices:
         interval_list (list[int]): Interval list for temporal augmentation.
             It will randomly pick an interval from interval_list and sample
             frame index with the interval.
-        filename_tmpl (str): Template for file name. Default: '{:08d}.png'.
     """
 
-    def __init__(self, interval_list, filename_tmpl='{:08d}.png'):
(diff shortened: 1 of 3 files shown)


mmedit/datasets/pipelines/augmentation.py
```diff
@@ -965,15 +963,18 @@ def __call__(self, results):
         # add the corresponding file paths
         lq_path_root = results['lq_path']
         gt_path_root = results['gt_path']
+
+        # retrieve the file names in the directory
+        filenames = sorted(
```
innerlee, line 968: * scandir should only do once. Otherwise disk access will become speed bottleneck
* This will fail if the backend is ceph.

So I changed my mind: support a start_index, instead of scanning folders

ckkelvinchan: Okay.

PR #341, opened 2021-05-29
[Feature] Add HourGlass

Yshuo-Li:
Part of DIC
HourGlass --> FeedbackHourGlass --> DIC

### DIFF
--- a/mmedit/models/extractors/feedback_hour_glass.py
+++ b/mmedit/models/extractors/feedback_hour_glass.py
@@ -61,7 +61,7 @@ class HourGlass(nn.Module):
     """
 
     def __init__(self, depth, mid_channels):
-        super(HourGlass, self).__init__()
+        super().__init__()
         self.up1 = ResBlock(mid_channels, mid_channels)
         self.pool = nn.MaxPool2d(2, 2)
         self.low1 = ResBlock(mid_channels, mid_channels)
--- a/mmedit/models/extractors/feedback_hour_glass.py
+++ b/mmedit/models/extractors/feedback_hour_glass.py
@@ -1,0 +1,90 @@
+import torch.nn as nn
+
+
+class ResBlock(nn.Module):
+    """ResBlock for HourGlass.
+
+    It has a style of:
+
+    ::
+
+        ---Conv-ReLU-Conv-Conv-+-
+         |_________Conv________|
+
+        or
+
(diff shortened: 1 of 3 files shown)


ckkelvinchan (approved):

mmedit/models/extractors/feedback_hour_glass.py
```diff
@@ -0,0 +1,90 @@
...
+    It is a recursive model.
+
+    Args:
+        depth (int): Depth of HourGlass, the number of recursions.
+        mid_channels (int): Number of channels in the intermediate features.
+    """
+
+    def __init__(self, depth, mid_channels):
+        super().__init__()
+        self.up1 = ResBlock(mid_channels, mid_channels)
```
innerlee, line 65: Is this an upsampling?

mmedit/models/extractors/feedback_hour_glass.py
```diff
@@ -0,0 +1,90 @@
...
+            Tensor: Forward results.
+        """
+
+        residual = self.conv_block(x)
+        if self.skip_layer:
+            x = self.skip_layer(x)
+        return x + residual
+
+
+class HourGlass(nn.Module):
```
innerlee, line 53: Any relationship with this? https://github.com/open-mmlab/mmpose/blob/master/mmpose/models/backbones/hourglass.py

Yshuo-Li: Refer to this file.

mmedit/models/extractors/feedback_hour_glass.py
Yshuo-Li: No

innerlee: Give it a better name :)

Yshuo-Li: In https://github.com/open-mmlab/mmpose/blob/master/mmpose/models/backbones/hourglass.py, it is called `up1`

innerlee: Digged a little about the history. mmpose's version comes from mmdet https://github.com/open-mmlab/mmdetection/pull/2840.

It is a bad naming. However since it has been existed in multiple mm-repos, let's keep it for now.

PR #349, opened 2021-06-01
[Doc] Update description for Vimeo90K

ckkelvinchan:
Added folder structure, and instructions to use `prepare_vimeo90k_dataset.py`

### DIFF
--- a/tools/data/super-resolution/vimeo90k/README.md
+++ b/tools/data/super-resolution/vimeo90k/README.md
@@ -15,4 +15,42 @@
 }
 ```
 
-Download from [here](http://toflow.csail.mit.edu/)
+The training and test datasets can be download from [here](http://toflow.csail.mit.edu/).
+
+The Vimeo90K dataset has a `clip/sequence/img` folder structure:
+```text
+├── GT/LQ
+│   ├── 001
+│   │   ├── 0001
+│   │   │   ├── im1.png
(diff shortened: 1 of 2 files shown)


tools/data/super-resolution/vimeo90k/README.md
```diff
@@ -15,4 +15,42 @@
...
+
+
+
+
+## Prepare the annotation files for Vimeo90K dataset
+
+To prepare the annotation file for training, you need to download the official training list path for Vimeo90K from the official website, and run the following command:
+
+```shell
+python tools/data/super-resolution/vimeo90k/preprocess_vimeo90k_dataset.py None None ./data/Vimeo90K/official_train_list.txt
```
innerlee, line 43: The two `None` looks strange.

ckkelvinchan: Since `gt_path` and `lq_path` are required in `prepare_vimeo90k_dataset.py`, but not used if `make-lmdb` is False. 

This is an (ugly) workaround. Another solution is to make this argument optional.

innerlee: I see. They should be optional

ckkelvinchan: Okay, I will modify it.

innerlee (approved):

PR #379, opened 2021-06-17
[Fix] Fix bug in restoration_video_inference.py

ckkelvinchan:
## Motivation
In [restoration_video_inference.py](https://github.com/ckkelvinchan/mmediting/blob/704eb2c614fb0bce40eab6a028be41bc93df7032/mmedit/apis/restoration_video_inference.py#L78), when using 
```
for i in range(0, data.size(1) - window_size):
```
a part of the input frames is omitted. This results in fewer output frames.

## Modification
This PR fixes the bug and replace the original command by 
```
for i in range(0, data.size(1) - 2 * (window_size // 2)):
```

### DIFF
--- a/mmedit/apis/restoration_video_inference.py
+++ b/mmedit/apis/restoration_video_inference.py
@@ -75,7 +75,7 @@ def restoration_video_inference(model, img_dir, window_size, start_idx,
         if window_size > 0:  # sliding window framework
             data = pad_sequence(data, window_size)
             result = []
-            for i in range(0, data.size(1) - 2 * window_size):
+            for i in range(0, data.size(1) - 2 * (window_size // 2)):
                 data_i = data[:, i:i + window_size]
                 result.append(model(lq=data_i, test_mode=True)['output'])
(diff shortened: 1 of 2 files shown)


innerlee:
Can this be tested?

ckkelvinchan:
> Can this be tested?

The unittests stucked at `Install CUDA`.

innerlee:
Do all vid models require cuda?

ckkelvinchan:
> Do all vid models require cuda?

Not really, and those apis for inference are actually not included in unittests.

innerlee:
Would you like to have a try? You can use a dummy model in the test (say, one layer, as long as the input/output format is good)

innerlee:
Our general philosophy is to test everything 😃

ckkelvinchan:
> Would you like to have a try? You can use a dummy model in the test (say, one layer, as long as the input/output format is good)

Sure, let me think about how to write a test for it.

innerlee (approved):

Issue #391, opened 2021-06-25
edvr in mmediting can't match edvrm_x4_8x4_600k_reds checkpoint?


xiab3369:

xiab3369:
edvr in mmediting can't match edvrm_x4_8x4_600k_reds checkpoint?

innerlee:
Please post the full command line, and the full error with stacktraces.

xiab3369:
for company confidentiality mechanism, I'm not convenient to take a screenshot. 
./tools/dist_test.sh configs/restorers/edvr/edvr_reds.py mmediting_pth/edvrm_x4_8x4_600k_reds_20200622-ba4a43e4.pth 4 --save-path ./result  --out test.pkl

problem 
missing keys in source state_dict: generator.conv_first.weight ...........................................

xiab3369:
basicvsr can match the checkpoint well

innerlee:
@Yshuo-Li Please check if the edvr code was touched recently

innerlee:
@xiab3369 The filename `edvr_reds.py` is not present in our repo. Please post the modification. For stacktraces, you can remove the private info such as usernames and file paths, and keep other info

xiab3369:
we just change the name of the file, not change edvr_net in backbone and config file content

xiab3369:
Can you share the full command line that accurately inference edvr network with edvrm_x4_8x4_600k_reds_20200622-ba4a43e4.pth

xiab3369:
basicVSR also exist the same problem

xiab3369:
The checkpoint contains 'module', we remove 'module' in pth file and successfully resolve the issue.

innerlee:
Which version of mmcv is used? The "module" should be stripped by mmcv automatically.

xiab3369:
mmcv-full 1.3.7

xiab3369:
It is automatically installed by mmediting on torch1.4

ckkelvinchan:
I just tried on my side. `edvrm_wotsa_x4_g8_600k_reds` indeed has the problem about `.module`. 

But `edvrm_wotsa_x4_8x4_600k_reds` and `basicvsr_reds4` do not have the issue.

innerlee:
@ckkelvinchan Is this caused by the publish code?

innerlee:
We may add tests for published checkpoints 🤔

ckkelvinchan:
> @ckkelvinchan Is this caused by the publish code?

[...]

PR #396, opened 2021-06-25
[Fix] Change the path to pre-trained EDVR-M

ckkelvinchan:
## Motivation
As discussed in #391, there are additional `.modules`'s  in `edvrm_x4_8x4_600k_reds_20200622-ba4a43e4.pth`. Therefore, the weights cannot be correctly loaded.

## Modification
We manually removed the unnecessary `.modules`. The path to the modified model is changed.


innerlee:
Any idea on the cause?

ckkelvinchan:
> Any idea on the cause?

No clue at the moment. It seems not the problem of `publish_model`. I guess `model = model.module` is not execute for some unknown reasons.

innerlee:
Would you mind digging into this problem?

ckkelvinchan:
> Would you mind digging into this problem?

Sure I will try

ckkelvinchan:
I tried training the model and there is no error in this current version. 

I checked the commit history and found that two EDVR-M models are uploaded in the same commit. It is kind of strange that only one has this problem. May need more time but not sure whether I can figure it out.

innerlee (approved):

PR #406, opened 2021-07-01
[Fix] Fix normalization in restoration_video_inference

ckkelvinchan:
## Motivation
Most of the video SR models in MMEditing take inputs with range [0, 1], but TDAN takes inputs with range [-0.5, 0.5]. Therefore, `Normalize` should be added to the pipeline in `restoration_video_inference.py`.

## Modification
`demo_pipeline` is added in the configuration files of each VSR model. The demo will look for `demo_pipeline`.

## Remark
An additional modification is that the output images are named according to `args.filename_tmpl` and `args.start_idx`.

### DIFF
--- a/mmedit/apis/restoration_video_inference.py
+++ b/mmedit/apis/restoration_video_inference.py
@@ -38,35 +38,8 @@ def restoration_video_inference(model, img_dir, window_size, start_idx,
 
     device = next(model.parameters()).device  # model device
 
-    # pipeline
-    test_pipeline = [
-        dict(
-            type='GenerateSegmentIndices',
-            interval_list=[1],
-            start_idx=start_idx,
-            filename_tmpl=filename_tmpl),
-        dict(
-            type='LoadImageFromFileList',
-            io_backend='disk',
-            key='lq',
-            channel_order='rgb'),
-        dict(type='RescaleToZeroOne', keys=['lq']),
-    ]
-
-    # add normalize if it is used in the model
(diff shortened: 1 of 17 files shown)


innerlee:
A somewhat meta question... Why not add a test pipeline in configs and directly load from there?

ckkelvinchan:
> A somewhat meta question... Why not add a test pipeline in configs and directly load from there?

As the pipeline for sliding-window framework (e.g.  `EDVR`, `TDAN`) is not quite suitable for inferencing a sequence (they have multiple LQs but a single GT).

innerlee:
But the test pipeline do not have to be the same as val pipeline

mmedit/apis/restoration_video_inference.py
```diff
@@ -38,35 +38,8 @@ def restoration_video_inference(model, img_dir, window_size, start_idx,
...
-            test_pipeline.append(op)
-
-    test_pipeline.extend([
-        dict(type='FramesToTensor', keys=['lq']),
-        dict(type='Collect', keys=['lq'], meta_keys=['lq_path', 'key'])
-    ])
-
     # build the data pipeline
-    test_pipeline = Compose(test_pipeline)
+    test_pipeline = Compose(model.cfg.demo_pipeline)
```
innerlee, line 42: This should fall back to test_pipeline if no demo_pipeline exist, and to val_pipeline if neither demo_pipeline nor test_pipeline exist.

innerlee (approved):

PR #415, opened 2021-07-07
Add Chinese doc

ckkelvinchan:


docs/switch_language.md
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
innerlee, line 1: is this a make file or markdown

ckkelvinchan: There is a mistake. Will update now

innerlee (approved):

Issue #395, opened 2021-06-25
Reading data from S3


jasoromir:
Hi! I was wondering if it is possible to read data from AWS S3 to train a model such as BasicVSR.

I have been reading the docs and all indicates it could be done by modifying my config file, and setting the backend option to "ceph" instead of the default "disk".

Then I sett the paths to:
lq_folder = "S3://my-bucket/my-dataset/lq_folder"
gt_folder = "S3://my-bucket/my-dataset/gt_folder"

But I can't manage to make it work.
Probably I am missing something. Could you provide me with some extra guidance? :)

innerlee:
Please post the full stacktrace.

I guess there are authentication errors

jasoromir:
When I execute the following command in Google Colab:

`!./tools/dist_train.sh ./demo_files/my_config_BasicVSR.py 1 `

I get the following trace:

```
usr/local/lib/python3.7/dist-packages/torch/distributed/launch.py:164: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
  "The module torch.distributed.launch is deprecated "
The module torch.distributed.launch is deprecated and going to be removed in future.Migrate to torch.distributed.run
WARNING:torch.distributed.run:--use_env is deprecated and will be removed in future releases.
 Please read local_rank from `os.environ('LOCAL_RANK')` instead.
INFO:torch.distributed.launcher.api:Starting elastic_operator with launch configs:
  entrypoint       : ./tools/train.py
  min_nodes        : 1
  max_nodes        : 1
  nproc_per_node   : 1
  run_id           : none
  rdzv_backend     : static
  rdzv_endpoint    : 127.0.0.1:29500
  rdzv_configs     : {'rank': 0, 'timeout': 900}
  max_restarts     : 3
  monitor_interval : 5
  log_dir          : None
  metrics_cfg      : {}

INFO:torch.distributed.elastic.agent.server.local_elastic_agent:log directory set to: /tmp/torchelastic_4y6bxj4d/none_z_uv4j_g
INFO:torch.distributed.elastic.agent.server.api:[default] starting workers for entrypoint: python3
INFO:torch.distributed.elastic.agent.server.api:[default] Rendezvous'ing worker group
/usr/local/lib/python3.7/dist-packages/torch/distributed/elastic/utils/store.py:53: FutureWarning: This is an experimental API and will be changed in future.
  "This is an experimental API and will be changed in future.", FutureWarning
[...]

PR #423, opened 2021-07-08
[Feature] Support loading annotation from file for video SR datasets

ckkelvinchan:
## Motivation
As raised in #395, some datasets use `glob` or `scandir` when loading annotations. These functions are not compatible with ceph. As a result, it is unable to train some models when ceph is used.

## Modifications
1. For `SRFolderMultipleGTDataset` and `SRFolderVideoDataset`, an argument `ann_file` is added. If it is provided, annotations will be loaded from the file.
2. For `SRVimeo90KMultipleGTDataset`, `glob` is removed, and the file paths are replaced by the standard Vimeo-90K file names.

### DIFF
--- a/mmedit/datasets/sr_folder_multiple_gt_dataset.py
+++ b/mmedit/datasets/sr_folder_multiple_gt_dataset.py
@@ -1,6 +1,8 @@
 import glob
 import os.path as osp
 
+import mmcv
+
 from .base_sr_dataset import BaseSRDataset
 from .registry import DATASETS
 
@@ -70,20 +72,20 @@ class SRFolderMultipleGTDataset(BaseSRDataset):
     def _load_annotations_from_file(self):
         data_infos = []
 
-        with open(self.ann_file, 'r') as fin:
-            for line in fin:
-                key, sequence_length = line.strip().split(' ')
-                if self.num_input_frames is None:
-                    num_input_frames = sequence_length
-                else:
-                    num_input_frames = self.num_input_frames
-                data_infos.append(
-                    dict(
-                        lq_path=self.lq_folder,
-                        gt_path=self.gt_folder,
-                        key=key,
-                        num_input_frames=int(num_input_frames),
-                        sequence_length=int(sequence_length)))
+        ann_list = mmcv.list_from_file(self.ann_file)
+        for ann in ann_list:
+            key, sequence_length = ann.strip().split(' ')
+            if self.num_input_frames is None:
+                num_input_frames = sequence_length
+            else:
+                num_input_frames = self.num_input_frames
+            data_infos.append(
+                dict(
+                    lq_path=self.lq_folder,
+                    gt_path=self.gt_folder,
+                    key=key,
+                    num_input_frames=int(num_input_frames),
+                    sequence_length=int(sequence_length)))
 
         return data_infos
 
--- a/mmedit/datasets/sr_folder_multiple_gt_dataset.py
+++ b/mmedit/datasets/sr_folder_multiple_gt_dataset.py
@@ -9,18 +9,35 @@ from .registry import DATASETS
 class SRFolderMultipleGTDataset(BaseSRDataset):
     """General dataset for video super resolution, used for recurrent networks.
 
-    It assumes all video clips under the root directory is used for training
-    or test.
-
     The dataset loads several LQ (Low-Quality) frames and GT (Ground-Truth)
     frames. Then it applies specified transforms and finally returns a dict
     containing paired data and other information.
+
+    This dataset takes an annotation file specifying the sequences used in
+    training or test. If no annotation file is provided, it assumes all video
+    sequences under the root directory is used for training or test.
+
+    In the annotation file (.txt), each line contains:
+
+        1. folder name;
+        2. number of frames in this sequence (in the same folder)
+
+    Examples:
+
+    ::
+
+        calendar 41
+        city 34
+        foliage 49
+        walk 47
 
     Args:
         lq_folder (str | :obj:`Path`): Path to a lq folder.
         gt_folder (str | :obj:`Path`): Path to a gt folder.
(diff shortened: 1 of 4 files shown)


mmedit/datasets/sr_folder_video_dataset.py
```diff
@@ -57,18 +78,43 @@ def __init__(self,
...
 
         self.data_infos = self.load_annotations()
 
+    def _load_annotations_from_file(self):
+        self.folders = {}
+        data_infos = []
+
+        with open(self.ann_file, 'r') as fin:
+            for line in fin:
+                key, max_frame_num = line.strip().split(' ')
```
innerlee, line 93: use rsplit, and restrict the split num to 2.

This way, spaces in paths are supported.

mmedit/datasets/sr_folder_video_dataset.py
```diff
@@ -57,18 +78,43 @@ def __init__(self,
...
+                sequence = key.split('/')[0]
```
innerlee, line 94: Do not hardcode path separator.

mmedit/datasets/sr_vimeo90k_multiple_gt_dataset.py
```diff
@@ -57,10 +56,14 @@ def load_annotations(self):
 
         data_infos = []
         for key in keys:
-            lq_paths = sorted(
-                glob.glob(osp.join(self.lq_folder, key, '*.png')))
-            gt_paths = sorted(
-                glob.glob(osp.join(self.gt_folder, key, '*.png')))
+            lq_paths = [
+                osp.join(self.lq_folder, key, f'im{i}.png')
```
innerlee, line 60: Unrelated to this pr, but the hardcoded png format is a headache also.

mmedit/datasets/sr_folder_video_dataset.py
ckkelvinchan: If using `.split(' ', 2)`, for `path 1/000.png 10`, it will become `['path', 1/000.png', '10']` Need to join it back if we want to support spaces in paths.

innerlee: the number is num of splits. So 1 split splits to 2 items

note the "r" in rsplit

ckkelvinchan: Oh I missed `rsplit`. It should work then.

mmedit/datasets/sr_vimeo90k_multiple_gt_dataset.py
ckkelvinchan: It is the official format of Vimeo-90K, I think it is reasonable to take this format (?)

innerlee: Not really. Our codebase is for developers, who will use their own format (whatever their dataset happens uses)

ckkelvinchan: This class is for Vimeo-90K only. `SRFolderVideoDataset` or `SRFolderMultipleGTDataset` should be used for general purposes.

Actually we could support `start_idx` and `filename_tmpl`. But this way cannot support general cases. For instance, it fails when the files are named `im01.png, im03.png, im05.png, im07.png, im09.png, img11.png, img13.png`.

mmedit/datasets/sr_folder_multiple_gt_dataset.py
```diff
@@ -48,13 +67,36 @@ def __init__(self,
 
         self.data_infos = self.load_annotations()
 
+    def _load_annotations_from_file(self):
+        data_infos = []
+
+        with open(self.ann_file, 'r') as fin:
+            for line in fin:
```
innerlee, line 74: use `mmcv.list_from_file`

mmedit/datasets/sr_folder_video_dataset.py
```diff
@@ -57,18 +78,43 @@ def __init__(self,
...
         self.filename_tmpl = filename_tmpl
         self.metric_average_mode = metric_average_mode
 
         self.data_infos = self.load_annotations()
 
+    def _load_annotations_from_file(self):
+        self.folders = {}
+        data_infos = []
+
+        with open(self.ann_file, 'r') as fin:
```
innerlee, line 91: use `list_from_file`

mmedit/datasets/sr_vimeo90k_multiple_gt_dataset.py
innerlee: Okay lets keep it as is for now.

mmedit/datasets/sr_vimeo90k_multiple_gt_dataset.py
```diff
@@ -57,10 +56,14 @@ def load_annotations(self):
...
+                for i in range(1, 8)
+            ]
+            gt_paths = [
+                osp.join(self.gt_folder, key, f'im{i}.png')
+                for i in range(1, 8)
```
innerlee, line 65: 8 is a magic number. Can it be made into a variable at least?

ckkelvinchan: Yes it can be done. I will update this in the afternoon.

mmedit/datasets/sr_folder_multiple_gt_dataset.py
```diff
@@ -48,13 +69,36 @@ def __init__(self,
...
         """Load annoations for the dataset.
 
         Returns:
             dict: Returned dict for LQ and GT pairs.
         """
 
+        if self.ann_file:
+            return self._load_annotations_from_file()
+
         sequences = sorted(glob.glob(osp.join(self.lq_folder, '*')))
```
innerlee, line 102: here, the usage of glob

innerlee (approved):

PR #425, opened 2021-07-11
[Improve] Refactor liif for pytorch2onnx

Yshuo-Li:
Create `generator`


mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+from mmedit.models.builder import build_backbone, build_component
+from mmedit.models.registry import BACKBONES
+from mmedit.utils import get_root_logger
+
+
+class LIIFNet(nn.Module):
+    """LIIF net for single image super-resolution.
+
+    Paper: Learning Continuous Image Representation with
+           Local Implicit Image Function
```
ckkelvinchan, line 16: Add `CVPR, 2021`

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+        super().__init__()
+
+        self.local_ensemble = local_ensemble
+        self.feat_unfold = feat_unfold
+        self.cell_decode = cell_decode
+        self.eval_bsize = eval_bsize
+
+        # model
+        self.encoder = build_backbone(encoder)
+        imnet_in_dim = self.encoder.mid_channels
```
ckkelvinchan, line 43: There may be a problem if `self.encoder` has no `mid_channels`.

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+        if self.feat_unfold:
+            imnet_in_dim *= 9
+        imnet_in_dim += 2  # attach coord
```
ckkelvinchan, line 46: coord -> coordinates

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+        self.local_ensemble = local_ensemble
+        self.feat_unfold = feat_unfold
+        self.cell_decode = cell_decode
+        self.eval_bsize = eval_bsize
+
+        # model
+        self.encoder = build_backbone(encoder)
+        imnet_in_dim = self.encoder.mid_channels
+        if self.feat_unfold:
+            imnet_in_dim *= 9
```
ckkelvinchan, line 45: What is the `9` referring to?

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+            imnet_in_dim += 2
+        imnet['in_dim'] = imnet_in_dim
+        self.imnet = build_component(imnet)
+
+    def forward(self, x, coord, cell, test_mode=False):
+        """Forward function.
+
+        Args:
+            x: input tensor.
+            coord (Tensor): coord tensor.
```
ckkelvinchan, line 57: coord -> coordinate

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+
+        Args:
+            feature (Tensor): encoded feature.
+            coord (Tensor): coord tensor, shape (BHW, 2).
+            cell (Tensor | None): cell tensor. Default: None.
+
+        Returns:
+            result (Tensor): (part of) output.
+        """
+        feat = feature
```
ckkelvinchan, line 88: Why not directly set the argument to `feat`?

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+            result = result[:, :, 0, :].permute(0, 2, 1)
+            return result
+
+        if self.feat_unfold:
+            feat = F.unfold(
+                feat, 3, padding=1).view(feat.shape[0], feat.shape[1] * 9,
+                                         feat.shape[2], feat.shape[3])
+
+        if self.local_ensemble:
+            vx_lst = [-1, 1]
```
ckkelvinchan, line 105: It is better to replace `vx`, `rx`, ... by some more meaningful names.

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+                    mid_tensor = torch.cat([mid_tensor, rel_cell], dim=-1)
+
+                bs, q = coord.shape[:2]
+                pred = self.imnet(mid_tensor.view(bs * q, -1)).view(bs, q, -1)
+                preds.append(pred)
+
+                area = torch.abs(rel_coord[:, :, 0] * rel_coord[:, :, 1])
+                areas.append(area + 1e-9)
+
+        tot_area = torch.stack(areas).sum(dim=0)
```
ckkelvinchan, line 154: `tot` means total? It is not easy to understanding the meaning

ckkelvinchan: `areas = areas[::-1]` should work.

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+            x (Tensor): Input tensor.
+            coord (Tensor): coord tensor.
+            cell (Tensor): cell tensor.
+
+        Returns:
+            pred (Tensor): output of model.
+        """
+        with torch.no_grad():
+            n = coord.shape[1]
+            ql = 0
```
ckkelvinchan, line 181: What does `ql` mean?

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+            preds = []
+            while ql < n:
+                qr = min(ql + self.eval_bsize, n)
```
ckkelvinchan, line 184: I think the definition of `qr` is unncessary. When `qr > n`, `coord[:, ql:qr, :]` is equal to `coord[:, ql:, :]`

mmedit/models/backbones/sr_backbones/liif_net.py
```diff
@@ -0,0 +1,259 @@
...
+                 eval_bsize=None):
+        super().__init__(
+            encoder=encoder,
+            imnet=imnet,
+            local_ensemble=local_ensemble,
+            feat_unfold=feat_unfold,
+            cell_decode=cell_decode,
+            eval_bsize=eval_bsize)
+
+        self.conv_first = self.encoder.conv_first
```
ckkelvinchan, line 239: I think it is better to directly construct the EDSR modules here? In the current design, when someone accidentally uses other networks which do not have `conv_first`, then `self.encoder.conv_first` would lead to an error.

mmedit/models/backbones/sr_backbones/liif_net.py
Yshuo-Li: self.encoder can be EDSR or RDN, all of them have `mid_channels`.

mmedit/models/backbones/sr_backbones/liif_net.py
Yshuo-Li: 3x3 unfold

mmedit/models/backbones/sr_backbones/liif_net.py
Yshuo-Li: This is LIIFEDSR, the backbone should be EDSR.

mmedit/models/backbones/sr_backbones/liif_net.py
Yshuo-Li: changed to left

mmedit/models/backbones/sr_backbones/liif_net.py
Yshuo-Li: define `right` will be better.

mmedit/models/backbones/sr_backbones/liif_net.py
ckkelvinchan: Then we should add a value error? Now there is a chance that users use some other networks that contain no `mid_channels`.

mmedit/models/backbones/sr_backbones/liif_net.py
ckkelvinchan: But the encoder is not necessarily EDSR. It may not be the best way to define EDSR here.

ckkelvinchan: There is a chance that the user (accidentally) uses other backbones in LIIFEDSR. Maybe add a value error here?

Yshuo-Li: He should use other backbones.

ckkelvinchan: If the user uses SRCNN as `encoder` in LIIFEDSR, there will be an error since SRCNN has no `conv_first. 

Of course, this is not allowed. Therefore, it is better to raise an error so that the user could know why does the error occur. We can raise an ValueError when the EDSR is not used. 

Similar for line 43.

Yshuo-Li: Of course, if the user uses SRCNN as an `encoder` in LIIFEDSR, there will be an error. But we do not need to raise an error, because `python` will raise the following error itselfs 

`AttributeError: LIIF: LIIFEDSR: 'SRCNN' object has no attribute 'mid_channels'`

or

`AttributeError: LIIF: LIIFEDSR: 'SRCNN' object has no attribute 'conv_first'`

Why do we repeat the errors?

ckkelvinchan: They don't know the main reason is "SRCNN should not be used". I think for some users, they may not be able to relate "SRCNN has no mid_channels" and " We do not accept SRCNN here".

Even if no error is raised, we should at least tell the users in the docstring that only EDSR and RDN can be used. (LIIFEDSR may be fine as EDSR is in the class name, but LIIFNet should state this fact).

Yshuo-Li: LIIFNet is the base class of LIIF Net, and in LIIFNet (liif_net.py, line 18~19):

    The subclasses should define `generator` with `encoder` and `imnet`,
        and overwrite the function `gen_feature`.

LIIFEDSR is a subclass of LIIF, and LIIFRDN is in PR #428 .

User can add LIIFSRCNN or others.

ckkelvinchan: Then LIIFSRCNN cannot inherit `__init__` from LIIFNet, as there is no `mid_channels`?

ckkelvinchan: In this case, why don’t we move `build_backbone` to the subclasses?

Yshuo-Li: Calculate imnet_in_dim is a common process, which should be placed in `LIIFNet`.

 `imnet_in_dim = self.encoder.mid_channels`

So we need to build the backbone in `LIIFNet`.

1st, seldom backbone does have `mid_channels`
2ed, if the user wants to use these backbones like SRCNN, they can init the subclass without inheriting `__init__` from LIIFNet.

mmedit/models/backbones/sr_backbones/liif_net.py
innerlee: Let's keep status quo and add helpful message when issues flood in.

innerlee (approved):

PR #453, opened 2021-07-27
Fix  typo in docs/demo.md, in Restoration (Image) part，cause config file  not found.

610265158:
I thought that there is a mistake in docs/demo.md, in the Restoration (Image) part. The config file 'configs/restorer/esrgan/esrgan_x4c64b23g32_1x16_400k_div2k.py'  is pointing to the err one, it should be  'configs/restorer/esrgan/esrgan_x4c64b23g32_g1_400k_div2k.py', right?


innerlee:
Thanks!

innerlee (approved):

PR #460, opened 2021-08-03
Add facial landmark pipeline

Yshuo-Li:


requirements/runtime.txt
```diff
@@ -1,3 +1,4 @@
+face-alignment
```
innerlee, line 1: Put this to optional since it is used by a small part of people

mmedit/datasets/pipelines/generate_assistant.py
```diff
@@ -1,3 +1,4 @@
+import face_alignment
```
innerlee, line 1: wrap it with try catch to make it optional dependency

mmedit/datasets/pipelines/generate_assistant.py
```diff
@@ -166,3 +171,139 @@ def __repr__(self):
...
+
+    def __repr__(self):
+        repr_str = (f'{self.__class__.__name__}: '
+                    f'image_key={self.image_key}'
+                    f'landmark_key={self.landmark_key}')
+        return repr_str
+
+
+@PIPELINES.register_module()
+class FacialFeaturesLocation:
```
innerlee, line 230: Rename to `GenerateFaceLocation`? It does not have any deep features

innerlee: Or `LocateFaces`?

PR #465, opened 2021-08-05
[Fix] torch.new_tensor will cause device error when pytorch>=1.7

nbei:


innerlee (approved):

innerlee:
Can this be tested (in cpu mode)?

nbei:
> Can this be tested (in cpu mode)?

Sure. The modification is to guarantee that two tensors are in the same device.

innerlee:
Okay then please add a test case for this ;) My understanding is that the current code is only tested on gpu mode, so it will slip through the ci

nbei:
> Okay then please add a test case for this ;) My understanding is that the current code is only tested on gpu mode, so it will slip through the ci

No. Current code has been tested on CPU. It will not report error, because the tensors will be put on CPU by default.

innerlee:
yeah your right

PR #495, opened 2021-08-18
Beautify doc notes

Yshuo-Li:


docs_zh-CN/conf.py
```diff
@@ -33,8 +33,8 @@
 ]
 
 autodoc_mock_imports = [
```
innerlee, line 35: No its the mock imports, but the extensions. Please double check the referenced pr

Yshuo-Li: done.

innerlee (approved):

PR #504, opened 2021-08-26
[Feature] Support random degradations during training

ckkelvinchan:
## Motivation
Recent works on face restoration or real-world super-resolution rely heavily on applying random degradations during training.

## Modification
This PR adds support to generate random degradations in the data processing pipeline. It currently supports
1. Random resize
2. Gaussian blur
3. Sinc filter
4. Gaussian noise
5. Poisson noise
6. JPEG compression
7. Random shuffling of degradations

This is a generic class. Supporting other degradations can be done by adding corresponding operations in this class.


mmedit/datasets/pipelines/blur_kernels.py
```diff
@@ -0,0 +1,519 @@
+# This code is referenced from BasicSR with modifications.
+# Reference: https://github.com/xinntao/BasicSR/blob/master/basicsr/data/degradations.py  # noqa
+import random
+
+import numpy as np
+from scipy import special
+
+np.seterr(divide='ignore', invalid='ignore')
```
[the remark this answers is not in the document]
ckkelvinchan: When using the sinc kernel, line 405 results in NaN owing to the division by zero. But that particular point is overwritten in line 411. This line is just used to avoid this warning.

mmedit/datasets/pipelines/blur_kernels.py
```diff
@@ -0,0 +1,519 @@
...
+    """
+
+    diag = np.array([[sig_x**2, 0], [0, sig_y**2]])
+    rot = np.array([[np.cos(theta), -np.sin(theta)],
+                    [np.sin(theta), np.cos(theta)]])
+
+    return np.dot(rot, np.dot(diag, rot.T))
+
+
+def mesh_grid(kernel_size):
```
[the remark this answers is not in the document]
ckkelvinchan: In fact it uses `np.meshgrid`. It just stacks the outputs and combines wraps them as a function.

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1031,3 +1031,86 @@ def __repr__(self):
...
+
+        Args:
+            results (dict): A dict containing the necessary information and
+                data for augmentation.
+
+        Returns:
+            dict: A dict with a key added/modified.
+        """
+
+        results[self.dst_key] = results[self.src_key].copy()
```
innerlee, line 1063: do you need a deep copy here?

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1031,3 +1031,86 @@ def __repr__(self):
         repr_str = self.__class__.__name__
         repr_str += (f'(keys={self.keys})')
         return repr_str
+
+
+@PIPELINES.register_module()
+class CopyValueFromKey:
```
innerlee, line 1037: The name `CopyValueFromKey` is okayish. Does the name come from other mm-repos? If it is completely new, how about a shorter name e.g. `Clone/Copy/CopyValues`?

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1031,3 +1031,86 @@ def __repr__(self):
...
+    """Copy the value of a source key to a destination key.
+
+    It does the following: results[dst_key] = results[src_key].
+
+    Args:
+        src_key (str): The source key.
```
innerlee, line 1043: Do you need to support specifying a list of keys in one go?

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1031,3 +1031,86 @@ def __repr__(self):
...
+
+    def __repr__(self):
+        repr_str = self.__class__.__name__
+        repr_str += (f'(src_key={self.src_key})')
+        repr_str += (f'(dst_key={self.dst_key})')
+        return repr_str
+
+
+@PIPELINES.register_module()
+class RoundClipZeroOne:
```
innerlee, line 1074: So this is a kind of quantization?

mmedit/datasets/pipelines/blur_kernels.py
```diff
@@ -0,0 +1,519 @@
+# This code is referenced from BasicSR with modifications.
+# Reference: https://github.com/xinntao/BasicSR/blob/master/basicsr/data/degradations.py  # noqa
```
innerlee, line 2: Add licence info of the reference repo

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1031,3 +1031,86 @@ def __repr__(self):
...
+        self.keys = keys
+
+    def _round_clip(self, input_):
+        if isinstance(input_, np.ndarray):
+            input_ = [input_]
+
+        # round and clip
+        input_ = [np.clip((v * 255.0).round(), 0, 255) / 255. for v in input_]
+
+        if len(input_) == 1:
```
innerlee, line 1093: This will cause trouble for len-1 input lists.

innerlee:
A general note: for augmentations, please add the added/modified keys in docstring.

mmedit/datasets/pipelines/blur_kernels.py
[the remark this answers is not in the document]
innerlee: Okay, it should be localized to that function. Please use np.geterr to store original settings and restore them at the end of the function.

mmedit/datasets/pipelines/blur_kernels.py
```diff
@@ -0,0 +1,519 @@
...
+        sigma_matrix (ndarray): The variance matrix with shape (2, 2).
+        grid (ndarray): Coordinates generated by :func:`mesh_grid`,
+            with shape (K, K, 2), where K is the kernel size.
+
+    Returns:
+        kernel (ndarrray): Un-normalized kernel.
+    """
+
+    inverse_sigma = np.linalg.inv(sigma_matrix)
+    kernel = np.exp(-0.5 * np.sum(np.dot(grid, inverse_sigma) * grid, 2))
```
innerlee, line 66: The behavior of np.dot is not very consistent see https://numpy.org/doc/stable/reference/generated/numpy.dot.html. So if you meant mat mult, then use matmul.

A guideline is to avoid using np.dot if possible

mmedit/datasets/pipelines/blur_kernels.py
```diff
@@ -0,0 +1,519 @@
...
+                                   sig_y,
+                                   theta,
+                                   beta,
+                                   grid=None,
+                                   is_isotropic=True):
+    """Generate a bivariate generalized Gaussian kernel.
+
+    Described in `Parameter Estimation For Multivariate Generalized
+    Gaussian Distributions` by Pascal et. al (2013). In isotropic mode,
+    only `sig_x` is used. `sig_y` and `theta` is ignored.
```
innerlee, line 121: same here

mmedit/datasets/pipelines/blur_kernels.py
```diff
@@ -0,0 +1,519 @@
...
+
+    Returns:
+        ndarray: Rotated sigma matrix.
+    """
+
+    diag = np.array([[sig_x**2, 0], [0, sig_y**2]])
+    rot = np.array([[np.cos(theta), -np.sin(theta)],
+                    [np.sin(theta), np.cos(theta)]])
+
+    return np.dot(rot, np.dot(diag, rot.T))
```
innerlee, line 27: Same here, avoid np.dot

mmedit/datasets/pipelines/blur_kernels.py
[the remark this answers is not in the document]
innerlee: Okay, rename it to `_mesh_grid`

mmedit/datasets/pipelines/blur_kernels.py
```diff
@@ -0,0 +1,519 @@
...
+
+    inverse_sigma = np.linalg.inv(sigma_matrix)
+    kernel = np.reciprocal(
+        np.power(np.sum(np.dot(grid, inverse_sigma) * grid, 2), beta) + 1)
+    kernel = kernel / np.sum(kernel)
+
+    return kernel
+
+
+def random_bivariate_gaussian(kernel_size,
```
innerlee, line 197: random_bivariate_gaussian -> random_bivariate_gaussian_kernel

innerlee: Same for other similar functions

mmedit/datasets/pipelines/augmentation.py
ckkelvinchan: No for now, but let us support that now in case we need that in the future.

mmedit/datasets/pipelines/augmentation.py
ckkelvinchan: Yes it is. After the degradations, the low-quality images are not quantized. This function just turns them to quantized images

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1031,3 +1031,97 @@ def __repr__(self):
...
+    Added keys are the keys in the attribute "dst_keys".
+
+    Args:
+        src_keys (list[str]): The source keys.
+        dst_keys (list[str]): The destination keys.
+    """
+
+    def __init__(self, src_keys, dst_keys):
+
+        if len(src_keys) != len(dst_keys):
```
innerlee, line 1053: add assertion that they are lists, in case someone pass a str

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1031,3 +1031,97 @@ def __repr__(self):
...
+
+        Args:
+            results (dict): A dict containing the necessary information and
+                data for augmentation.
+
+        Returns:
+            dict: A dict with a key added/modified.
+        """
+        for (src_key, dst_key) in zip(self.src_keys, self.dst_keys):
+            results[dst_key] = results[src_key].copy()
```
innerlee, line 1071: use deepcopy

mmedit/datasets/pipelines/augmentation.py
innerlee: Okay the name "round clip" is not very clear. It is not simply a clip after a round. How about naming it as "Quantize"?

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1031,3 +1031,97 @@ def __repr__(self):
...
+        repr_str += (f'(dst_keys={self.dst_keys})')
+        return repr_str
+
+
+@PIPELINES.register_module()
+class RoundClipZeroOne:
+    """Round off and clip the image to [0, 1].
+
+    It is assumed that the the input has range [0, 1].
+
```
innerlee, line 1087: added/modified keys

mmedit/datasets/pipelines/blur_kernels.py
```diff
@@ -0,0 +1,530 @@
...
+    """Generate a bivariate isotropic or anisotropic Gaussian kernel.
+
+    In isotropic mode, only `sig_x` is used. `sig_y` and `theta` are
+    ignored.
+
+    Args:
+        kernel_size (int): The size of the kernel
+        sig_x (float): Standard deviation along horizontal direction
+        sig_y (float | None, optional): Standard deviation along the vertical
+            direction. If it is None, 'is_isotropic' must be set to True.
```
innerlee, line 85: Add an assertion for this

mmedit/datasets/pipelines/blur_kernels.py
```diff
@@ -0,0 +1,530 @@
...
+
+    Returns:
+        ndarray: Rotated sigma matrix.
+    """
+
+    diag = np.array([[sig_x**2, 0], [0, sig_y**2]])
+    rot = np.array([[np.cos(theta), -np.sin(theta)],
+                    [np.sin(theta), np.cos(theta)]])
+
+    return np.matmul(rot, np.matmul(diag, rot.T))
```
innerlee, line 26: this mat will be float64.

innerlee: A general comment: operations in this file seem to be all float64. But it is slower than float32. Our convention is to use float32 for pipelines if possible. Float64 is used only when the precision is critical for that op.

One simple way to fix this is to add a line of note in the docstring that this op is computed in double precision.
A better fix is to check if some of them are valid in float32, and use float32 for those

image

mmedit/datasets/pipelines/random_degradations.py
```diff
@@ -0,0 +1,388 @@
...
+
+from mmedit.datasets.pipelines import blur_kernels as blur_kernels
+from ..registry import PIPELINES
+
+
+@PIPELINES.register_module()
+class RandomBlur:
+    """Apply random blur to the input.
+
+    Args:
```
innerlee, line 14: As mentioned in previous comment, please add the affected keys

innerlee: for all augmentations

mmedit/datasets/pipelines/random_degradations.py
```diff
@@ -0,0 +1,388 @@
+import random
```
innerlee, line 1: use np.random

mmedit/datasets/pipelines/random_degradations.py
```diff
@@ -0,0 +1,388 @@
...
+        params (dict): A dictionary specifying the degradation settings.
+        keys (list[str]): A list specifying the keys whose values are
+            modified.
+    """
+
+    def __init__(self, params, keys):
+        self.keys = keys
+        self.params = params
+
+    def _apply_gaussian_noise(self, input_):
```
innerlee, line 185: How about `input_` -> `input`?

innerlee (approved):

PR #507, opened 2021-08-27
[Feature] Add MATLAB-like bicubic interpolation

ckkelvinchan:
## Motivation
Existing image and video SR methods commonly use the `imresize` function in MATLAB to generate LR images. It means that one would need MATLAB to generate such LR images.

## Modification
This PR adds the Python version of the MATLAB `imresize` function.

**Note**:
There is a slight difference between the official MATLAB implementation and this implementation. For instance, when using BasicVSR on REDS4, we achieve the following results:

- Official MATLAB implementation: 31.4170/0.8909
- This implementation: 31.4206/0.8911

### DIFF
--- a/mmedit/datasets/pipelines/matlab_resize.py
+++ b/mmedit/datasets/pipelines/matlab_resize.py
@@ -2,6 +2,8 @@ from math import ceil
 
 import numpy as np
 
+from ..registry import PIPELINES
+
 
 def get_size_from_scale(input_size, scale_factor):
     """Get the output size given input size and scale factor.
@@ -52,7 +54,7 @@ def cubic(x):
 
     """
 
-    x = np.array(x).astype(np.float64)
+    x = np.array(x).astype(np.float32)
     x_abs = np.absolute(x)
     x_abs_sq = np.multiply(x_abs, x_abs)
     x_abs_cu = np.multiply(x_abs_sq, x_abs)
@@ -76,7 +78,7 @@ def get_weights_indices(input_length, output_length, scale, kernel,
         kernel_width (int): The width of the kernel.
 
     Returns:
-        list[ndarry]: The weights and the indices for interpolation.
+        list[ndarray]: The weights and the indices for interpolation.
 
 
     """
@@ -91,7 +93,7 @@ def get_weights_indices(input_length, output_length, scale, kernel,
         kernel_width = kernel_width
 
     # coordinates of output
-    x = np.arange(1, output_length + 1).astype(np.float64)
+    x = np.arange(1, output_length + 1).astype(np.float32)
 
     # coordinates of input
     u = x / scale + 0.5 * (1 - 1 / scale)
@@ -134,7 +136,7 @@ def resize_along_dim(img_in, weights, indices, dim):
         ndarray: Interpolated (along one dimension) image.
     """
 
-    img_in = img_in.astype(np.float64)
+    img_in = img_in.astype(np.float32)
     w_shape = weights.shape
     output_shape = list(img_in.shape)
     output_shape[dim] = w_shape[0]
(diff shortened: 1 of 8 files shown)


mmedit/datasets/pipelines/matlab_resize.py
```diff
@@ -0,0 +1,270 @@
...
+    def __init__(self,
+                 keys,
+                 scale=None,
+                 output_shape=None,
+                 kernel='bicubic',
+                 kernel_width=4.0):
+
+        if kernel.lower() != 'bicubic':
+            raise ValueError('Currently support bicubic kernel only.')
+        else:
```
innerlee, line 198: needless `else`

mmedit/datasets/pipelines/matlab_resize.py
```diff
@@ -0,0 +1,270 @@
...
+    if img_in.dtype == np.uint8:
+        img_out = np.clip(img_out, 0, 255)
+        return np.around(img_out).astype(np.uint8)
+    else:
+        return img_out
+
+
+@PIPELINES.register_module()
+class MATLABResize:
+    """Resize the input image using MATLAB downsampling.
```
innerlee, line 169: `Resize the input image using MATLAB-like downsampling.`

mmedit/datasets/pipelines/matlab_resize.py
```diff
@@ -0,0 +1,270 @@
...
+
+        Currently support bicubic interpolation only. Note that the output of
+        this function is slightly different from the official MATLAB function.
+
```
innerlee, line 173: Give a note on the added/modified keys like others https://github.com/open-mmlab/mmediting/blob/master/mmedit/datasets/pipelines/augmentation.py#L26-L28

innerlee:
One important thing is to be explicit about whether the implementation is the "same" as matlab's. From my point of view, this is a `matlab-like resize`. So some phrases in the docstring should be adjusted slightly to void giving a wrong impression.

innerlee:
Also please rebase master to pass CI

mmedit/datasets/pipelines/matlab_resize.py
```diff
@@ -0,0 +1,270 @@
+from math import ceil
```
innerlee, line 1: Use np.ceil

mmedit/datasets/pipelines/matlab_resize.py
```diff
@@ -0,0 +1,270 @@
...
+
+    scale = [
+        1.0 * output_shape / input_shape
+        for (input_shape, output_shape) in zip(input_size, output_size)
+    ]
+
+    return scale
+
+
+def cubic(x):
```
innerlee, line 46: name: `_cubic`

mmedit/datasets/pipelines/matlab_resize.py
```diff
@@ -0,0 +1,270 @@
...
+    Args:
+        x (ndarray): The distance from the center position.
+
+    Returns:
+        ndarray: The weight corresponding to a particular distance.
+
+    """
+
+    x = np.array(x).astype(np.float32)
+    x_abs = np.absolute(x)
```
innerlee, line 58: Not essential, but the coding style is verbose. Try use simpler notations

```python
  x_abs = np.abs(x)
  x_abs_sq = x_abs * x_abs
  x_abs_cu = x_abs_sq * x_abs
```

mmedit/datasets/pipelines/matlab_resize.py
```diff
@@ -0,0 +1,270 @@
...
+    x_abs_sq = np.multiply(x_abs, x_abs)
+    x_abs_cu = np.multiply(x_abs_sq, x_abs)
+    f = np.multiply(1.5 * x_abs_cu - 2.5 * x_abs_sq + 1,
```
innerlee, line 61: Add a line of note about the math formula used in L61-65

𝑦=1.5𝑥³−2.5𝑥²+1 if |𝑥| <= 1, ...

mmedit/datasets/pipelines/matlab_resize.py
```diff
@@ -0,0 +1,270 @@
...
+    left = np.floor(u - kernel_width / 2)  # leftmost pixel
+    p = int(ceil(kernel_width)) + 2  # maximum number of pixels
+
+    # indices of input pixels
+    ind = np.expand_dims(left, axis=1) + np.arange(p) - 1
+    indices = ind.astype(np.int32)
+
+    # weights of input pixels
+    weights = h(np.expand_dims(u, axis=1) - indices - 1)
+    weights = np.divide(weights,
```
innerlee, line 109: same here, 

* use `/` to replace np.divide
* use None to replace np.expand_dims

innerlee:
And give licence info if the code is adapted from elsewhere

innerlee (approved):

innerlee (approved):

PR #529, opened 2021-09-10
[Feature] Helper function for face restoration

ckkelvinchan:
## Motivation
Existing face restoration methods generally detect and align the faces, perform restoration, and then paste back to the original image. Therefore, we need such a helper function for that.

## Modification
This PR adds a helper function to 1) detect faces, 2) warp and crop faces, and 3) paste the restored faces back to the input image.


ckkelvinchan:
Will use another library. Closing this PR.

PR #543, opened 2021-09-18
Fix test.py and integrate 'eval_result'

Yshuo-Li:
## Motivation
When using `test.py`, we do not get eval results even though `eval_result` is contained in the dictionary. That is caused by #532 , since `outputs` is a list of dictionaries, and `eval_result` will not be contained in `outputs`.

## Modification
This PR fixes the bug caused by #532 and changes `eval_results` to `eval_result` in some models to ensure `dataset.evaluate` is bypassed if `eval_result` is not contained in the outputs.


ckkelvinchan (approved):

configs/inpainting/deepfillv1/deepfillv1_256x256_4x4_celeba.py
```diff
@@ -146,7 +146,7 @@
         test_mode=True),
     test=dict(
         type=dataset_type,
-        ann_file=(f'{data_root}/val_celeba_img_list.txt'),
+        ann_file=('work_dirs/val_celeba_img_list copy.txt'),
```
innerlee, line 149: This path is not portable

innerlee:
@Yshuo-Li Any info on the motivation? Which bug is fixed?

ckkelvinchan:
> @Yshuo-Li Any info on the motivation? Which bug is fixed?

Fixing #532. maybe @Yshuo-Li can add this info.

innerlee:
> maybe @Yshuo-Li can add this info.

Yeah prs should come with detailed info. Otherwise it is clueless and informal. We discourage this

innerlee (approved):

PR #553, opened 2021-09-27
Fix `update_model_index` and add a hook in pre-commit to update metafile automatically.

Yshuo-Li:
## Motivation
- Fix Metadata.Name in `update_model_index` 
- Add a hook in pre-commit to update metafile automatically

## Modification
- Metadata.Name in `update_model_index` 
- Add a hook in `.pre-commit-config.yaml`


innerlee:
Linting fails

innerlee (approved):

PR #557, opened 2021-09-28
[Feature] Unsharp masking

ckkelvinchan:
## Motivation
The recent blind SR method Real-ESRGAN adopts unsharp masking during training to improve the visual quality.

## Modification
This PR supports unsharp masking in the processing pipeline.

### DIFF
--- a/mmedit/datasets/pipelines/augmentation.py
+++ b/mmedit/datasets/pipelines/augmentation.py
@@ -1131,3 +1131,67 @@ class Quantize:
 
     def __repr__(self):
         return self.__class__.__name__
+
+
+@PIPELINES.register_module()
+class UnsharpMasking:
+    """Apply unsharp masking to an image or a sequence of images.
+
+    Args:
+        kernel_size (int): The kernel_size of the Gaussian kernel.
+        sigma (float): The standard deviation of the Gaussian.
+        weight (float): The weight of the "details" in the final output.
+        threshold (float): Pixel differences larger than this value are
+            regarded as "details".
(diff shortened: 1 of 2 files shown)


mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1131,3 +1131,67 @@ def __call__(self, results):
...
+            regarded as "details".
+        keys (list[str]): The keys whose values are processed.
+
+    Added keys are "xxx_unsharp", where "xxx" are the attributes specified
+    in "keys".
+
+    """
+
+    def __init__(self, kernel_size, sigma, weight, threshold, keys):
+        if kernel_size % 2 == 0:
```
innerlee, line 1154: why not throw an error?

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -1131,3 +1131,67 @@ def __call__(self, results):
...
+
+    def _unsharp_masking(self, imgs):
+        is_single_image = False
+        if isinstance(imgs, np.ndarray):
+            is_single_image = True
+            imgs = [imgs]
+
+        outputs = []
+        for img in imgs:
+            residue = img - cv2.filter2D(img, -1, self.kernel)
```
innerlee, line 1174: what's the type of img? if it is uint8, then is this `-` correct?

ckkelvinchan: In our pipeline, the images should be loaded as a float. I think it is fine?

PR #545, opened 2021-09-22
Update config and pipeline of DIC

Yshuo-Li:
## Motivation
The prior pipeline of DIC loads facial landmarks from files while generating from images by face_alignment will be better.

## Modification
- Add GenerateLandmark
- Update config and pipeline of DIC

PR #571, opened 2021-10-12
[Fix] Update DIC

Yshuo-Li:
Update config and README of DIC, follow #545 

The reason: 
- There is a difference between the old version and the ori repo, although the results of the old version of DIC is no less than that of the ori repo. (The align loss in the old version is different from the ori repo)
- In the old version, we load landmark from a pre-generated file. Now we can generate landmark in pipeline.


ckkelvinchan:
> > > Update config and README of DIC, follow #545
> > 
> > 
> > Please specify why you update it.
> 
> * There is a difference between the old version and the ori repo, although the results of the old version of DIC is no less than that of the ori repo. (The align loss in the old version is different from the ori repo)
> * In the old version, we load landmark from a pre-generated file. Now we can generate landmark in pipeline.

Which files correspond to the changes? I saw only README and metafile are changed.

PR #612, opened 2021-11-12
[Fix] Convert minimum mmcv version to 5.4

LeoXing1996:


ckkelvinchan:
@LeoXing1996 It should be gcc version instead of mmcv?

LeoXing1996:
> @LeoXing1996 It should be gcc version instead of mmcv?

Yes, compiling mmcv needs GCC version larger than 5.4.

ckkelvinchan:
> > @LeoXing1996 It should be gcc version instead of mmcv?
> 
> Yes, compiling mmcv needs GCC version larger than 5.4.

Okay~

PR #622, opened 2021-11-23
[Feature] Support video input and output in restoration demo

ckkelvinchan:
## Motivation
When inferencing a video model, one would need to convert the video (e.g., `.mp4`) to images, and then load the images. This causes inconvenience if one wants to quickly try the result.

## Modification
This PR supports using video format as input and output. One could simply replace the directory with the path to the video.

**Note**: When saving the outputs as a video, it introduces video compression, which would lower the visual quality.


demo/restoration_video_demo.py
```diff
@@ -28,26 +33,51 @@ def parse_args():
...
+        help='maximum sequence length if recurrent framework is used')
     parser.add_argument('--device', type=int, default=0, help='CUDA device id')
     args = parser.parse_args()
     return args
 
 
 def main():
+    """ Demo for video restoration models.
+
+    Note that we accept video as input(output), when 'input_dir'('output_dir')
```
Yshuo-Li, line 49: input (output)
'input_dir' ('output_dir')

demo/restoration_video_demo.py
```diff
@@ -28,26 +33,51 @@ def parse_args():
...
+    is set to the path to the video. But using videos introduces video
+    compression, which lower the visual quality. If you want actual quality,
```
Yshuo-Li, line 51: which lowers ...

demo/restoration_video_demo.py
```diff
@@ -28,26 +33,51 @@ def parse_args():
...
+    please save them as separate images (.png).
```
Yshuo-Li, line 52: please save them as separated images (.png).

Yshuo-Li (commented):
Several lines are not covered by tests

ckkelvinchan:
> Several lines are not covered by tests

It is tested in GPU only. Therefore it is not revealed in the CI report.

Yshuo-Li (approved):

Yshuo-Li (approved):

PR #626, opened 2021-11-23
[Fix] Change 'radius' to 'kernel_size' for UnsharpMasking in Real-ESRNet config

ckkelvinchan:


huoshuai-dot:
kernel_size must be odd num,kernel_size=15?

ckkelvinchan:
> kernel_size must be odd num,kernel_size=15?

It should be 51 according to the official implementation. Modified in this PR too.

PR #628, opened 2021-11-24
[Doc] Add Abstract to README (restorer)

ckkelvinchan:


nbei:
We should also add a teaser for these methods.

ckkelvinchan:
> We should also add a teaser for these methods.

Okay got it. Will do it now.

PR #633, opened 2021-11-30
[Feature] Add RealBasicVSR model

ckkelvinchan:
Paper: [Investigating Tradeoffs in Real-World Video Super-Resolution](https://arxiv.org/pdf/2111.12704.pdf)


nbei:
You may fix the unit test.

ckkelvinchan:
> You may fix the unit test.

Yes, will do~

mmedit/models/restorers/real_basicvsr.py
```diff
@@ -0,0 +1,190 @@
...
+
+from ..builder import build_loss
+from ..common import set_requires_grad
+from ..registry import MODELS
+from .real_esrgan import RealESRGAN
+
+
+@MODELS.register_module()
+class RealBasicVSR(RealESRGAN):
+    """RealBasicVSR model for single image super-resolution.
```
Yshuo-Li, line 13: Video super-resolution

mmedit/models/restorers/real_basicvsr.py
```diff
@@ -0,0 +1,190 @@
...
+
+        Returns:
+            dict: Returned output.
+        """
+
+        # during initialization, load weights from the ema model
+        if (self.step_counter == self.start_iter
+                and self.generator_ema is not None):
+            if is_module_wrapper(self.generator):
+                self.generator.module.load_state_dict(
```
Yshuo-Li, line 90: Please consider this situation in the unit test

ckkelvinchan: The first `if` is considered. For the second one, it is True only when distributed/multiple GPU learning is used, and therefore I think it is not testable here.

PR #650, opened 2021-12-09
[CI] Cancel previous runs that are not completed

wangruohui:
Follow https://github.com/open-mmlab/mmcv/pull/1537


wangruohui:
As a test, [this action](https://github.com/open-mmlab/mmediting/actions/runs/1558522784) is canceled after a new commit.

ckkelvinchan (approved):

PR #660, opened 2021-12-16
[Enhance]: Use mmcv root model registry.

RangiLyu:
## Motivation

MMEditing did not use mmcv's MODELS registry as root/parent registry. This PR makes it impossible to build models across projects.

## Modification

mmpose/models/registry.py

```Python
from mmcv.cnn import MODELS as MMCV_MODELS
from mmcv.utils import Registry

MODELS = Registry('model', parent=MMCV_MODELS)
BACKBONES = MODELS
COMPONENTS = MODELS
LOSSES = MODELS

MODELS = Registry(
    'models', build_func=build_model_from_cfg, parent=MMCV_MODELS)
```


wangruohui (approved):
This modification will merge all three registries into a single one, which follows standard of other repos do.

PR #676, opened 2021-12-20
[Feature] Add model inpaintor and discriminator for AOT-GAN

quincylin1:
added model for aot-gan


mmedit/models/components/discriminators/smpatch_disc.py
```diff
@@ -0,0 +1,128 @@
...
+                bias=False,
+                norm_cfg=norm_cfg,
+                act_cfg=dict(type='LeakyReLU', negative_slope=0.2),
+                with_spectral_norm=with_spectral_norm)
+        ]
+
+        # stacked intermediate layers,
+        # gradually increasing the number of filters
+        multiple_now = 1
+        multiple_prev = 1
```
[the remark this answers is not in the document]
ckkelvinchan: Maybe use `multiplier_in` and `multiplier_out`?

mmedit/models/components/discriminators/smpatch_disc.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # output one-channel prediction map
+        sequence += [
+            build_conv_layer(
+                dict(type='Conv2d'),
+                base_channels * multiple_now,
+                1,
+                kernel_size=kernel_size,
+                stride=1,
+                padding=padding)
```
ckkelvinchan, line 96: So here is a conventional `nn.Conv2d`?

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,279 @@
...
+from mmedit.core import tensor2img
+from ..common import set_requires_grad
+from ..registry import MODELS
+from .one_stage import OneStageInpaintor
+
+
+@MODELS.register_module()
+class AOTInpaintor(OneStageInpaintor):
+    """Inpaintor for AOT-GAN method.
+    This inpaintor is implemented according to the paper:
```
ckkelvinchan, line 16: Empty line

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,279 @@
...
+    Aggregated Contextual Transformations for High-Resolution Image Inpainting
+    Args:
```
ckkelvinchan, line 18: Empty line

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,279 @@
...
+        super().__init__(
+            encdec,
+            disc=disc,
+            loss_gan=loss_gan,
+            loss_composed_percep=loss_composed_percep,
+            loss_out_percep=loss_out_percep,
+            loss_l1_valid=loss_l1_valid,
+            train_cfg=train_cfg,
+            test_cfg=test_cfg,
+            pretrained=pretrained)
```
ckkelvinchan, line 53: If it is identical to its parent class, you can remove this part.

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def forward_train_d(self, data_batch, is_real, is_disc, **kwargs):
+        """Forward function in discriminator training step.
+        In this function, we compute the prediction for each data batch (real
```
ckkelvinchan, line 57: Empty line

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,279 @@
...
+                training step. Otherwise, this function is called in generator
+                training step. This will help us to compute different types of
+                adversarial loss, like LSGAN.
+            **kwargs (keyword arguments).
+        Returns:
+            dict: Contains the loss items computed in this function.
+        """
+
+        pred = self.disc(data_batch)
+        loss_ = self.loss_gan(pred, is_real, is_disc, **kwargs)
```
ckkelvinchan, line 75: Is `**kwargs` necessary here?

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,279 @@
...
+        loss = dict(real_loss=loss_) if is_real else dict(fake_loss=loss_)
+
+        if self.with_disc_shift_loss:
+            loss_d_shift = self.loss_disc_shift(loss_)
+            # 0.5 for average the fake and real data
+            loss.update(loss_disc_shift=loss_d_shift * 0.5)
+
+        return loss
+
+    def generator_loss(self, fake_res, fake_img, data_batch):
```
ckkelvinchan, line 86: This should be identical to that in `OneStageInpaintor`?

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,279 @@
...
+
+        res = dict(
+            gt_img=gt.cpu(),
+            masked_img=masked_img.cpu(),
+            fake_res=fake_res.cpu(),
+            fake_img=fake_img.cpu())
+
+        return res, loss
+
+    def forward_test(self,
```
ckkelvinchan, line 132: This should be identical to that in `OneStageInpaintor`?

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,279 @@
...
+                [masked_img,
+                 mask.expand_as(masked_img), fake_res, fake_img])
+            img = torch.cat(img_list, dim=3).cpu()
+            self.save_visualization(img, osp.join(save_path, filename))
+            output['save_img_path'] = osp.abspath(
+                osp.join(save_path, filename))
+
+        return output
+
+    def train_step(self, data_batch, optimizer):
```
ckkelvinchan, line 202: This should be identical to that in `OneStageInpaintor`?

ckkelvinchan (commented):
Hi @quincylin1, it seems that some of the functions are overlapped with the one in `OneStageInpaintor`? May I ask which ones are different?

mmedit/models/components/discriminators/smpatch_disc.py
quincylin1: yes

mmedit/models/inpaintors/aot_inpaintor.py
quincylin1: I overwrote the generator_loss function in aot inpaintor because there are a few functions within it which take different arguments from the parent onestageinpaintor. 

For example, in aot inpaintor the l1_valid_loss is not weighted by the mask, so weight is not passed into the function self.loss_l1_valid:

image

whereas in onestageinpaintor the l1 valid loss is weighted by the mask:

image

mmedit/models/inpaintors/aot_inpaintor.py
quincylin1: again there are a few subtle differences. For example, the holes of `masked_img` for aot inpaintor are filled with ones by combining with mask:

image

whereas `masked_img` for onestageinpaintor has holes filled with zeros.

Another difference is the `data_dict` used for calculating the metrics:

For aot inpaintor:

image

For onestageinpaintor:

image

mmedit/models/inpaintors/aot_inpaintor.py
quincylin1: Again there are a few subtle differences, for example `self.train_cfg.disc_step` is not used in aot inpaintor. Please compare the two train_step functions for onestageinpaintor and aot_inpaintor for more details.

mmedit/models/inpaintors/aot_inpaintor.py
ckkelvinchan: Oh okay, I missed that. Thanks for your clarification.

mmedit/models/components/discriminators/smpatch_disc.py
ckkelvinchan: Then I think we can directly use `nn.Conv2d`.

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,252 @@
...
+        fake_res = self.generator(input_x)
+        fake_img = gt_img * (1. - mask) + fake_res * mask
+
+        # reconstruction losses
+        results, g_losses = self.generator_loss(fake_res, fake_img, data_batch)
+        loss_g, log_vars_g = self.parse_losses(g_losses)
+        log_vars.update(log_vars_g)
+
+        # adversarial loss
+        set_requires_grad(self.disc, True)
```
ckkelvinchan, line 210: I do not see `set_requires_grad(self.disc, False)`. Is that normal? As I see line 405 [here](self.disc_step_count)

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,252 @@
...
+        fake_data = fake_img.detach()
+        real_data = gt_img
+        disc_losses_real = self.forward_train_d(
+            real_data, True, True, mask=mask)
```
ckkelvinchan, line 214: `real_data = gt_img` is not necessary. `self.forward_train_d(gt_img, True, True, mask=mask)` suffices.  Similar for `fake_data`.

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,252 @@
...
+        disc_losses_fake = self.forward_train_d(
+            fake_data, False, True, mask=mask)
+        disc_losses_ = disc_losses_real['real_loss'] + disc_losses_fake[
+            'fake_loss']
+        disc_losses = dict(disc_losses=disc_losses_)
+        print(disc_losses)
```
ckkelvinchan, line 220: Remove

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,252 @@
...
+        set_requires_grad(self.disc, True)
+        fake_data = fake_img.detach()
+        real_data = gt_img
+        disc_losses_real = self.forward_train_d(
+            real_data, True, True, mask=mask)
+        disc_losses_fake = self.forward_train_d(
+            fake_data, False, True, mask=mask)
+        disc_losses_ = disc_losses_real['real_loss'] + disc_losses_fake[
+            'fake_loss']
+        disc_losses = dict(disc_losses=disc_losses_)
```
ckkelvinchan, line 219: We usually separate the real_loss and fake_loss. You can refer to [here](https://github.com/open-mmlab/mmediting/blob/037efba0c80e40cb3a0f94b6a17e4621412db520/mmedit/models/inpaintors/one_stage.py#L363) for the implementation.

quincylin1: the official implementation combined the real and fake losses together and then back propagate the total disc loss. Wonder if this would affect the results if we split the real and fake loss and back propagate each individually.

mmedit/models/inpaintors/aot_inpaintor.py
quincylin1: I think the reason line 405 sets `set_requires_grad(self.disc, False)` is to reduce the computational overhead when calculating losses for generator. If you look at line 364 `set_requires_grad(self.disc, True)` when the discriminator is trained

mmedit/models/inpaintors/aot_inpaintor.py
ckkelvinchan: I think it won't make a difference. But in any case we can follow the official implementation.

mmedit/models/inpaintors/aot_inpaintor.py
ckkelvinchan: If we do not set it to False, the discriminator will also have gradient when optimizing the generator. That could lead to incorrect training.

quincylin1: I see, I guess I'll follow the `onestageinpaintor` then.

quincylin1: Just a sanity check, I have screenshot the official code below. Would be great if you could also compare my `aot_inpaintor` with the official implementation to see if there's any major difference between the two. Thanks!

image

I have included the log part below too for your reference:

image

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,253 @@
...
+
+        loss = dict()
+
+        if self.with_gan:
+            pred = self.disc(fake_img)
+            loss_g_fake = self.loss_gan(pred, True, False, mask=mask)
+            loss['loss_g_fake'] = loss_g_fake
+
+        if self.with_l1_valid_loss:
+            loss_loss_l1_valid = self.loss_l1_valid(fake_res, gt)
```
ckkelvinchan, line 86: `loss_l1_valid`

mmedit/models/inpaintors/aot_inpaintor.py
ckkelvinchan: In the official code, adversarial loss is computed using `comp_img`, but the output of the generator is used for computing adversarial loss here?

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,253 @@
...
+                image. Defaults to False.
+            save_path (str, optional): If given a valid str, the reuslts will
+                be saved in this path. Defaults to None.
+            iteration (int, optional): Iteration number. Defaults to None.
+
+        Returns:
+            dict: Contain output results and eval metrics (if exist).
+        """
+
+        masked_img = masked_img.float() + mask
```
ckkelvinchan, line 126: If I am not mistaken, the only difference to `forward_test` in `OneStageInpaintor` is this line. 

I think we can do the following instead of copying the entire function:
```
masked_img = masked_img.float() + mask
super().forward_test(masked_img, mask, save_image, save_path, iteration, **kwargs)

```

ckkelvinchan: Or could we modify the data pipeline instead? Then `forward_test` and `train_step` here can follow `OneStageInpaintor`.

mmedit/models/inpaintors/aot_inpaintor.py
quincylin1: The output of the generator here is `fake_res`, whereas `comp_img` in the official code corresponds to `fake_img` here. 

Screenshot from 2022-01-20 09-57-45

`fake_img` is then used to compute the adversarial loss, which is the same input as `comp_img` in the official code:

Screenshot from 2022-01-20 09-57-56

mmedit/models/inpaintors/aot_inpaintor.py
quincylin1: In the official code, the ssim and psnr metric are not weighted by the mask, so I also set `mask=None` in `forward_test` in aot_inpaintor here:

Screenshot from 2022-01-20 10-06-11

whereas in `OneStageInpaintor`, `mask` is not set to None when calculating psnr and ssim in `forward_test`

Screenshot from 2022-01-20 10-08-32

mmedit/models/inpaintors/aot_inpaintor.py
ckkelvinchan: Oh sorry! I missed that. Then I think there is no problem.

mmedit/models/inpaintors/aot_inpaintor.py
ckkelvinchan: Okay then we can keep it as it is. But is it common that the holes are filled with different values? As you mentioned that some methods fill the holes with 1 and some fill with 0. It may be a good features to support such a choices when getting `masked_img` in the data pipeline.

quincylin1: emm im not sure.. I think its only the AOT-GAN method that fills the holes with ones using `masked_img = masked_img.float() + mask` (maybe I am wrong, need to double check that). We may be able to support such a choice by adding an additional `fill_ones` parameter in `GetMaskedImage` in `mmedit.datasets.pipelines.formating.py`?

ckkelvinchan: Yes I think so. Instead of using `fill_ones`, I think it is better for users to specify the mask value (within 0 and 1). Then 1 - mask will represent valid regions.

Are you able to help with this?

quincylin1: Is it necessary though, since most masks are simply binary mask?

ckkelvinchan: I was thinking about this too. But since fill_ones is a special case of supporting arbitrary values and it does not complicate the code much, I think it may be better to support a more general use case, in case some future methods have something to do with the mask (but it may eventually be useless haha)

quincylin1: i see! But are we specifying the value of the hole region or the value of the mask itself? The latter is usually loaded from some files.

ckkelvinchan: Then that would complicate the situation a bit.

i am currently away from home, I will think about it and get back to you soon. Thanks!

quincylin1: No worries! Thanks kelvin!

quincylin1: @ckkelvinchan any update on this?

ckkelvinchan: Sorry, I forgot to reply to this thread. 

I think in this case we can use a simple argument `is_invert_mask`. If True, then we will swap the value of 1 and 0. Just forget the other values for now~

quincylin1: emmm interesting. In our case the mask is not inverted (i.e. the masked region is still ones) but that `masked_img` is added to `mask` so as to fill the holes with ones. Wonder if `is_invert_mask` would do the trick here

ckkelvinchan: Okay I know what you mean. Then let's forget the `is_invert_mask`, and keep it as it is.

mmedit/models/inpaintors/aot_inpaintor.py
```diff
@@ -0,0 +1,253 @@
...
+        1. get fake res/image
+        2. compute reconstruction losses for generator
+        3. compute adversarial loss for discriminator
+        4. optimize generator
+        5. optimize discriminator
+
+        Args:
+            data_batch (torch.Tensor): Batch of data as input.
+            optimizer (dict[torch.optim.Optimizer]): Dict with optimizers for
+                generator and discriminator (if have).
```
ckkelvinchan, line 190: if exist

ckkelvinchan (commented):
One minor comment, overall LGTM :)

ckkelvinchan:
> One minor comment, overall LGTM :)

Please start adding the unittest, thanks!

quincylin1:
> > One minor comment, overall LGTM :)
> 
> Please start adding the unittest, thanks!

@ckkelvinchan just added unittest, but not sure why the coverage for aot_inpaintor is low. When i run on my local machine the coverage is 97% for aot_inpaintor, with only line 49-51 and 94 missing. 
Screenshot from 2022-01-26 15-48-21

ckkelvinchan:
> > > One minor comment, overall LGTM :)
> > 
> > 
> > Please start adding the unittest, thanks!
> 
> @ckkelvinchan just added unittest, but not sure why the coverage for aot_inpaintor is low. When i run on my local machine the coverage is 97% for aot_inpaintor, with only line 49-51 and 94 missing.
> 
> Screenshot from 2022-01-26 15-48-21

Since the coverage here is based on CPU only. I saw in your code that most of the code is under the condition `if torch.cuda.is_available():`. Therefore it is not "covered". You can copy a similar code and test it on CPU.

quincylin1:
> > > > One minor comment, overall LGTM :)
> > > 
> > > 
> > > Please start adding the unittest, thanks!
> > 
> > 
> > @ckkelvinchan just added unittest, but not sure why the coverage for aot_inpaintor is low. When i run on my local machine the coverage is 97% for aot_inpaintor, with only line 49-51 and 94 missing.
> > Screenshot from 2022-01-26 15-48-21
> 
> Since the coverage here is based on CPU only. I saw in your code that most of the code is under the condition `if torch.cuda.is_available():`. Therefore it is not "covered". You can copy a similar code and test it on CPU.

@ckkelvinchan all checks have been passed, please check

ckkelvinchan (approved):

ckkelvinchan:
@quincylin1 Thank you! I will then review the next PR.

PR #677, opened 2021-12-20
[Feature] Add losses and metrics for AOT-GAN

quincylin1:
added losses and metrics for aot gan


ckkelvinchan:
Sorry for the delay. I was on leave last week. I will provide comments soon!

mmedit/models/losses/perceptual_loss.py
```diff
@@ -119,7 +119,8 @@ class PerceptualLoss(nn.Module):
     """
 
     def __init__(self,
-                 layer_weights,
+                 layer_weights_perceptual,
+                 layer_weights_style,
```
ckkelvinchan, line 123: That would affect all models using perceptual loss, leading to a several compatibility issue. 

A workaround is to keep `layer_weights` for perceptual loss. We can then add `layer_weights_style`. By default it is None and when it is not None, we initialize `self.vgg_style`. The computations in `forward` are then based on whether `self.vgg_style` exists.

mmedit/models/losses/perceptual_loss.py
```diff
@@ -160,26 +167,31 @@ def forward(self, x, gt):
...
                 percep_loss += self.criterion(
-                    x_features[k], gt_features[k]) * self.layer_weights[k]
+                    x_features_perceptual[k], gt_features_perceptual[k]
+                ) * self.layer_weights_perceptual[k]
             percep_loss *= self.perceptual_weight
         else:
             percep_loss = None
 
+        x_features_style = self.vgg_style(x)
+        gt_features_style = self.vgg_style(gt.detach())
```
ckkelvinchan, line 185: Move these lines under line 188

mmedit/models/losses/gan_loss.py
```diff
@@ -83,20 +88,189 @@ def forward(self, input, target_is_real, is_disc=False):
...
                 loss = self.loss(1 + input).mean()
             else:  # for generators in hinge-gan
                 loss = -input.mean()
+        elif self.gan_type == 'smgan':
+            mask = kwargs['mask']
+
+            input_height, input_width = input.shape[2:]
+            mask_height, mask_width = mask.shape[2:]
+
+            # Handle inconsistent size between outputs and masks
```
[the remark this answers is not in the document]
quincylin1: @ckkelvinchan If I remember correctly, it is for handling the case when the `RandomResizeCrop` function in data augmentation yields a result with size which is off by 1 pixel (e.g. sometimes the shape of result is (511, 512) rather than exactly (512, 512))

mmedit/models/losses/gan_loss.py
```diff
@@ -83,20 +88,189 @@ def forward(self, input, target_is_real, is_disc=False):
...
+                'Input x type is not a torch.Tensor. Got {}'.format(type(x)))
+        if not len(x.shape) == 4:
+            raise ValueError(
+                'Invalid input shape, we expect BxCxHxW. Got: {}'.format(
+                    x.shape))
+        _, c, _, _ = x.shape
+        tmp_kernel = self.kernel.to(x.device).to(x.dtype)
+        kernel = tmp_kernel.repeat(c, 1, 1, 1)
+
+        return conv2d(x, kernel, padding=self.padding, stride=1, groups=c)
```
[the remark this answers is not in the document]
quincylin1: I'm using torch v1.5.0 with torchvision v0.6.0a0+82fd1c8 here, but `[torchvision.transforms.GaussianBlur](https://pytorch.org/vision/main/generated/torchvision.transforms.GaussianBlur.html)` is only added for torch v1.6.0 and beyond.

mmedit/models/losses/gan_loss.py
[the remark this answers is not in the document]
ckkelvinchan: Why is that? Is it a bug or expected output?

mmedit/models/losses/gan_loss.py
```diff
@@ -83,20 +88,189 @@ def forward(self, input, target_is_real, is_disc=False):
...
+
         target_label = self.get_target_label(input, target_is_real)
         if self.gan_type == 'hinge':
             if is_disc:  # for discriminators in hinge-gan
                 input = -input if target_is_real else input
                 loss = self.loss(1 + input).mean()
             else:  # for generators in hinge-gan
                 loss = -input.mean()
+        elif self.gan_type == 'smgan':
+            mask = kwargs['mask']
```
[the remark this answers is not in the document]
ckkelvinchan: Since SMGAN is a LSGAN with mask altered, I think we can shift this part back to the AOT-GAN itself to keep `gan_loss.py` clean.

mmedit/models/losses/gan_loss.py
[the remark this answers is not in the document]
ckkelvinchan: Oh okay, following the suggestions in `gan_loss.py`, I think we can shift this part to AOT-GAN model. This function should be put under `gan_loss.py`.

mmedit/models/losses/gan_loss.py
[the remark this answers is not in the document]
quincylin1: but SMGAN also calculates the loss differently. Specifically, if `is_disc=False`, `loss = self.loss(input, target_label) * mask / mask.mean()` (see line 122-123), so it is not just that the labels are different.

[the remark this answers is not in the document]
ckkelvinchan: Okay. Maybe we can keep it as it is for now.

mmedit/models/losses/gan_loss.py
```diff
@@ -84,20 +89,189 @@ def forward(self, input, target_is_real, is_disc=False, mask=None):
...
+
+    Shape:
+        - input: Tensor with shape of (n, c, h, w)
+        - output: Tensor with shape of (n, c, h, w)
+
+    Examples:
+
+      >>> input = torch.rand(2, 4, 5, 5)
+      >>> gauss = kornia.filters.GaussianBlur((3, 3), (1.5, 1.5))
+      >>> output = gauss(input) # 2 x 4 x 5 x 5
```
ckkelvinchan, line 157: This should be irrelevant to this code. Or you should remove `kornia.filters.`

mmedit/models/losses/gan_loss.py
```diff
@@ -26,13 +28,15 @@ def __init__(self,
                  loss_weight=1.0):
         super().__init__()
         self.gan_type = gan_type
-        self.loss_weight = loss_weight
         self.real_label_val = real_label_val
         self.fake_label_val = fake_label_val
+        self.loss_weight = loss_weight
+
+        self.gaussian_blur = GaussianBlur()
```
ckkelvinchan, line 35: create this only with `smgan`

mmedit/models/losses/gan_loss.py
```diff
@@ -72,7 +77,7 @@ def get_target_label(self, input, target_is_real):
             self.real_label_val if target_is_real else self.fake_label_val)
         return input.new_ones(input.size()) * target_val
 
-    def forward(self, input, target_is_real, is_disc=False, mask=None):
+    def forward(self, input, target_is_real, is_disc=False, **kwargs):
```
ckkelvinchan, line 80: Why do you change `mask` to `**kwargs`?

mmedit/models/losses/perceptual_loss.py
```diff
@@ -167,26 +182,36 @@ def forward(self, x, gt):
...
 
         # calculate perceptual loss
         if self.perceptual_weight > 0:
             percep_loss = 0
-            for k in x_features.keys():
+            for k in list(self.layer_weights.keys()):
                 percep_loss += self.criterion(
-                    x_features[k], gt_features[k]) * self.layer_weights[k]
+                    x_features_perceptual[k],
+                    gt_features_perceptual[k]) * self.layer_weights[k]
```
ckkelvinchan, line 194: I suggest to keep this part unchanged.

mmedit/models/losses/perceptual_loss.py
```diff
@@ -132,16 +136,27 @@ def __init__(self,
...
+        self.layer_weights = layer_weights
         self.norm_img = norm_img
         self.perceptual_weight = perceptual_weight
         self.style_weight = style_weight
-        self.layer_weights = layer_weights
-        self.vgg = PerceptualVGG(
-            layer_name_list=list(layer_weights.keys()),
+        self.layer_weights_style = layer_weights_style
+
+        self.vgg_perceptual = PerceptualVGG(
```
ckkelvinchan, line 145: simply using `self.vgg` is fine.

mmedit/models/losses/perceptual_loss.py
```diff
@@ -132,16 +136,27 @@ def __init__(self,
...
             pretrained=pretrained)
 
+        if self.layer_weights_style is None:
+            self.layer_weights_style = self.layer_weights
+        else:
+            self.vgg_style = PerceptualVGG(
+                layer_name_list=list(self.layer_weights_style.keys()),
+                vgg_type=vgg_type,
+                use_input_norm=use_input_norm,
+                pretrained=pretrained)
```
ckkelvinchan, line 158:
```
if self.layer_weights_style is not None and self.layer_weights_style != self.layer_weights:
    self.vgg_style = PerceptualVGG(
        layer_name_list=list(self.layer_weights_style.keys()),
        vgg_type=vgg_type,
        use_input_norm=use_input_norm,
        pretrained=pretrained)
else:
    self.vgg_style = None
```

mmedit/models/losses/perceptual_loss.py
```diff
@@ -132,16 +136,27 @@ def __init__(self,
...
                  criterion='l1'):
         super().__init__()
+        self.layer_weights = layer_weights
         self.norm_img = norm_img
         self.perceptual_weight = perceptual_weight
         self.style_weight = style_weight
-        self.layer_weights = layer_weights
-        self.vgg = PerceptualVGG(
-            layer_name_list=list(layer_weights.keys()),
+        self.layer_weights_style = layer_weights_style
```
ckkelvinchan, line 143: Not necessary

mmedit/models/losses/perceptual_loss.py
```diff
@@ -167,26 +182,36 @@ def forward(self, x, gt):
...
 
         # calculate style loss
         if self.style_weight > 0:
+            if self.layer_weights_style != self.layer_weights:
+                x_features_style = self.vgg_style(x)
+                gt_features_style = self.vgg_style(gt.detach())
+
+            else:
+                x_features_style = x_features_perceptual
+                gt_features_style = gt_features_perceptual
```
ckkelvinchan, line 207: We can simply do this:
```
if self.vgg_style is not None:
    x_features = self.vgg_style(x)
    gt_features = self.vgg_style(gt.detach())
```

Then we will use the original code.

ckkelvinchan (commented):
Thanks!

mmedit/models/losses/perceptual_loss.py
```diff
@@ -136,12 +140,25 @@ def __init__(self,
         self.perceptual_weight = perceptual_weight
         self.style_weight = style_weight
         self.layer_weights = layer_weights
+        self.layer_weights_style = layer_weights_style
```
ckkelvinchan, line 143: This line is not necessary. In line 151 we can use `layer_weights_style` instead of `self.layer_weights_style`.

ckkelvinchan (commented):
Just two minor points. We can start writing the unit-test. Thanks!

mmedit/models/losses/perceptual_loss.py
quincylin1: Hi @ckkelvinchan sorry for the late reply! I have just been assigned to a different project so have been rather busy with it. 
Since `layer_weights_style` is also called in the `forward` function, I think keeping it as `self.layer_weights_style` may make life easier. As you can see in the `forward` function, `layer_weights_style` becomes undefined when `self` is removed from `__init__()`

image

ckkelvinchan: Don’t be sorry, and we greatly appreciate your help!

I missed that part and yes we should keep this. Thanks for your reminder. In this case I think there is only one minor point (as mentioned above) and we are good to go for the unittest. Really thanks for your help!

quincylin1: It's really been a while haha. Cool let me write up the unit tests asap! Thanks kelvin!

ckkelvinchan:
@quincylin1 It seems strange that only 1.6 failed. Could you help have a check?

image

quincylin1:
> @quincylin1 It seems strange that only 1.6 failed. Could you help have a check?
> 
> image

Yes I am looking into this too, not sure what's gone wrong yet but will keep checking

ckkelvinchan:
> > @quincylin1 It seems strange that only 1.6 failed. Could you help have a check?
> > image
> 
> Yes I am looking into this too, not sure what's gone wrong yet but will keep checking

Thanks!

quincylin1:
> > > @quincylin1 It seems strange that only 1.6 failed. Could you help have a check?
> > > image
> > 
> > 
> > Yes I am looking into this too, not sure what's gone wrong yet but will keep checking
> 
> Thanks!

image

quincylin1:
> > > > @quincylin1 It seems strange that only 1.6 failed. Could you help have a check?
> > > > image
> > > 
> > > 
> > > Yes I am looking into this too, not sure what's gone wrong yet but will keep checking
> > 
> > 
> > Thanks!
> 
> image

@ckkelvinchan Its interesting when I run the tests on my local machine I get a number that is different from the one here from the 7th decimal place onwards. Wonder if it is to do with float32 accuracy

quincylin1:
> > > > > @quincylin1 It seems strange that only 1.6 failed. Could you help have a check?
> > > > > image
> > > > 
> > > > 
> > > > Yes I am looking into this too, not sure what's gone wrong yet but will keep checking
> > > 
> > > 
> > > Thanks!
> > 
> > 
> > image
> 
> @ckkelvinchan Its interesting when I run the tests on my local machine I get a number that is different from the one here from the 7th decimal place onwards. Wonder if it is to do with float32 accuracy

We can change it to `npt.assert_almost_equal(loss.item(), 3.786323, decimal=6)` to avoid the 7th decimal (the default option for `assert_almost_equal`, what do you think?

ckkelvinchan:
> > > > > > @quincylin1 It seems strange that only 1.6 failed. Could you help have a check?
> > > > > > image
> > > > > 
> > > > > 
> > > > > Yes I am looking into this too, not sure what's gone wrong yet but will keep checking
> > > > 
> > > > 
> > > > Thanks!
> > > 
> > > 
> > > image
> > 
> > 
> > @ckkelvinchan Its interesting when I run the tests on my local machine I get a number that is different from the one here from the 7th decimal place onwards. Wonder if it is to do with float32 accuracy
> 
> We can change it to `npt.assert_almost_equal(loss.item(), 3.786323, decimal=6)` to avoid the 7th decimal (the default option for `assert_almost_equal`, what do you think?

Yeah I think you are right. Let's change it to `decimal=6`.

ckkelvinchan (approved):
Thanks @quincylin1 !

PR #688, opened 2021-12-28
[Feature] Add video interpolation demo

Yshuo-Li:


demo/video_interpolation_demo.py
```diff
@@ -0,0 +1,84 @@
...
+    parser.add_argument('config', help='test config file path')
+    parser.add_argument('checkpoint', help='checkpoint file')
+    parser.add_argument('input_dir', help='directory of the input video')
+    parser.add_argument('output_dir', help='directory of the output video')
+    parser.add_argument(
+        '--fps',
+        type=int,
+        default=50,
+        help='frames-per-second in the output video, which is needed when'
+        'the input is image sequence and want to get an output video.')
```
ckkelvinchan, line 25: frame rate of the output video, which is needed when the input is an image sequence and a video is desired as output.

ckkelvinchan: Another comment: Since in frame interpolation, we usually reduce the frame rate according to the number of frames we interpolate. Do you think it is better to use a multiplier (e.g. `0.5, 0.25`), and the ouptut frame rate equals to the this number times the original frame rate?

demo/video_interpolation_demo.py
```diff
@@ -0,0 +1,84 @@
...
+
+    model = init_model(
+        args.config, args.checkpoint, device=torch.device('cuda', args.device))
+
+    output, fps = video_interpolation_inference(model, args.input_dir,
+                                                args.start_idx,
+                                                args.max_seq_len,
+                                                args.batch_size)
+
+    fps = fps if fps > 0 else args.fps
```
ckkelvinchan, line 65: The following use case is not feasible: 

I have a video, I want to do interpolation and reduce the frame rate for slow motion.

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,122 @@
...
+    img_bytes = FILE_CLIENT.get(filepath)
+    image = mmcv.imfrombytes(
+        img_bytes, flag='color', channel_order='rgb', backend='pillow')
+    return image
+
+
+def video_interpolation_inference(model,
+                                  input_dir,
+                                  start_idx=0,
+                                  end_idx=None,
```
ckkelvinchan, line 36: It is not used in `video_interpolation_demo`?

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Args:
+        model (nn.Module): The loaded model.
+        input_dir (str): Directory of the input video.
+        start_idx (int): The index corresponds to the first frame in the
+            sequence. Default: 0.
+        end_idx (int | None): The index corresponds to the last interpolated
+            frame in the sequence. If it is None, interpolate to the last
+            frame of video or sequence. Default: None.
+        batch_size (int): Batch size. Default: 4.
```
ckkelvinchan, line 48: Although it is not a major issue, it's better to be consistent with the default value in `demo`.

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Returns:
+        output (list[numpy.array]): The predicted interpolation result.
+            It is an image sequence: [ori, pred, ori, pred, ori, ...]
```
ckkelvinchan, line 52: Some algorithms can interpolate more than one frame.

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,122 @@
...
+        files = files[start_idx:end_idx]
+        images = [read_image(f) for f in files]
+
+    data = []
+    for i in range(len(images) - 1):
+        data.append(
+            dict(
+                inputs=[images[i], images[i + 1]],
+                inputs_path=None,
+                key=input_dir))
```
ckkelvinchan, line 90: Some methods, such as QVI, take 4 frames as input and outputs the intermediate frames. Therefore, it is not good to hard-code it like that.

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,122 @@
...
+    data = [test_pipeline(d) for d in data]
+    data = scatter(collate(data, samples_per_gpu=1), [device])[0]['inputs']
+
+    # forward the model
+    result = [np.flip(images[0], axis=2)]
+    index = 1
+    with torch.no_grad():
+        length = data.shape[0]
+        for i in range(0, length, batch_size):
+            start, end = i, i + batch_size
```
ckkelvinchan, line 113:
```
for start in range(0, length, batch_size):
  end = start + batch_size
```

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,122 @@
...
+            output = model(data[start:end], test_mode=True)['output'].cpu()
+            for j in range(output.shape[0]):
+                new_image = tensor2img(output[j])
+                new_image = tensor2img(output[j])
```
ckkelvinchan, line 117: repeated

demo/video_interpolation_demo.py
```diff
@@ -0,0 +1,94 @@
...
+
+    output, fps = video_interpolation_inference(model, args.input_dir,
+                                                args.start_idx, args.end_idx,
+                                                args.batch_size)
+
+    if args.multiply_fps:
+        assert fps > 0, 'multiply_fps=True but the input is not a video'
+        fps = args.fps * fps
+    else:
+        fps = args.fps if args.fps > 0 else fps
```
ckkelvinchan, line 75: If we do it in this way, `args.fps` will have two meanings. This is not the ideal implementation. I suggest using `fps_multiplier` to directly calculate the output fps, if it is specified.

```
if args.fps_multiplier > 0:
  assert fps > 0, 'The input is not a video'
  fps = fps *= args.fps_multiplier
else:
  fps = args.fps if args.fps > 0 else fps
```

demo/video_interpolation_demo.py
```diff
@@ -0,0 +1,94 @@
...
+
+    file_extension = os.path.splitext(args.output_dir)[1]
+    if file_extension in VIDEO_EXTENSIONS:  # save as video
+        h, w = output[0].shape[:2]
+        fourcc = cv2.VideoWriter_fourcc(*'mp4v')
+        video_writer = cv2.VideoWriter(args.output_dir, fourcc, fps, (w, h))
+        for img in output:
+            print(img.shape)
```
ckkelvinchan, line 83: Remove this

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # check if the input is a video
+    input_fps = 0.0
+    file_extension = os.path.splitext(input_dir)[1]
+    if file_extension in VIDEO_EXTENSIONS:
+        video_reader = mmcv.VideoReader(input_dir)
+        input_fps = video_reader.fps
+        images = []
+        # load the images
+        for img in video_reader[start_idx:end_idx]:
```
ckkelvinchan, line 77: The use of `start_idx` does not match the description in the docstring. In the docstring, it means "if the first frame is 00000001.png, then start_idx=1"?

I remembered this is used for custom sequences whose index of the first frame is not zero.

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,140 @@
...
+        files = os.listdir(input_dir)
+        files = [osp.join(input_dir, f) for f in files]
+        files.sort()
+        files = files[start_idx:end_idx]
+        images = [read_image(f) for f in files]
+
+    data = []
+    before, after = io_sequence[0], io_sequence[1]
+
+    for i in range(0, len(images) - before - after + 1, before):
```
ckkelvinchan, line 89: Let say `before=after=2`. Assume interpolating one frame, then we should do 
1. `0, 1, 2, 3` -> `1.5`
2. `1, 2, 3, 4` -> `2.5`

But it seems that the code is doing
1. `0, 1, 2, 3` -> `1.5`
2. `2, 3, 4, 5` -> `3.5`
 
The interpolation between 1 and 2 is missing.

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,140 @@
...
+            images.append(np.flip(img, axis=2))  # BGR --> RGB
+    else:
+        files = os.listdir(input_dir)
+        files = [osp.join(input_dir, f) for f in files]
+        files.sort()
+        files = files[start_idx:end_idx]
+        images = [read_image(f) for f in files]
+
+    data = []
+    before, after = io_sequence[0], io_sequence[1]
```
ckkelvinchan, line 87: I think the model actually does not need the argument `io_sequence`. Maybe just take this as an argument of the demo. Just like `window_size` in `restoration_video_demo`.

mmedit/apis/video_interpolation_inference.py
ckkelvinchan: Moreover, does it cover the case when the network takes the entire sequence in and outputs the entire sequence? (Not sure if there exist such methods now)

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,140 @@
...
+                if len(output.shape) == 4:
+                    new_image = tensor2img(output[j])
+                    result.append(new_image)
+                else:
+                    for k in range(output.shape[1]):
+                        new_image = tensor2img(output[j][k])
+                        result.append(new_image)
+                if index + before + after >= len(images):
+                    result.extend(images[index + before:])
+                    break
```
ckkelvinchan, line 135: Similar to the above comment, there could be problems if `before` and `after` are not 1?

mmedit/apis/video_interpolation_inference.py
Yshuo-Li: If we do this, users must master the corresponding methods before using the demo

Yshuo-Li:
How about generating the sequence in each method?

ckkelvinchan:
> How about generating the sequence in each method?

I think the best would be to incorporate it into the data pipeline. But I am not sure whether it is easy to accomplish.

mmedit/models/video_interpolators/basic_interpolator.py
```diff
@@ -268,17 +269,18 @@ def split_frames(input_tensors):
...
 
-        num_frames = input_tensors.shape[0]
+        num_frames = input_tensors.shape[1]
 
-        result = []
-        for i in range(0, num_frames - 1, 1):
-            result.append(input_tensors[i:i + 2])
+        result = input_tensors[:, :2]  # the first tensor
+        for i in range(1, num_frames - 1, 1):
+            result = torch.cat([result, input_tensors[:, i:i + 2]], dim=0)
```
ckkelvinchan, line 283: Here `torch.cat` is called multiple times. I suggest we do the following:
```
result = [input_tensors[:, i:i + 2] for i in range(0, num_frames - 1)]
result = torch.cat(result, dim=0)
```

mmedit/apis/video_interpolation_inference.py
```diff
@@ -121,30 +110,26 @@ def video_interpolation_inference(model,
...
-    generate_frames = getattr(model.cfg['model'], 'generate_frames',
-                              default_generate_frames)
-    result = generate_frames(input_images, output_images)
+            output = model(data[start:end], test_mode=True)['output']
+            if len(output.shape) == 4:
+                output = output.unsqueeze(1)
+            if output_tensors is None:
+                output_tensors = output
+            else:
+                output_tensors = torch.cat([output_tensors, output], dim=0)
```
ckkelvinchan, line 131: Similarly, I think we can first put them in a list, and `cat` at last?

demo/video_interpolation_demo.py
```diff
@@ -0,0 +1,98 @@
...
+    args = parse_args()
+
+    model = init_model(
+        args.config, args.checkpoint, device=torch.device('cuda', args.device))
+
+    output, fps = video_interpolation_inference(model, args.input_dir,
+                                                args.start_idx, args.end_idx,
+                                                args.batch_size)
+
+    if args.fps_multiplier:
```
ckkelvinchan, line 76: There could be a problem if negative numbers are used.

mmedit/models/video_interpolators/basic_interpolator.py
Yshuo-Li: Good idea, thank you

demo/video_interpolation_demo.py
```diff
@@ -0,0 +1,99 @@
...
+        assert args.fps_multiplier > 0, '`fps_multiplier` can not be negative'
```
ckkelvinchan, line 77: cannot

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,134 @@
...
+                                  input_dir,
+                                  start_idx=0,
+                                  end_idx=None,
+                                  batch_size=4):
+    """Inference image with the model.
+
+    Args:
+        model (nn.Module): The loaded model.
+        input_dir (str): Directory of the input video.
+        start_idx (int): The index corresponds to the first frame in the
```
ckkelvinchan, line 58: The index corresponding

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,134 @@
...
+    Args:
+        input_images (list[np.array]): The input frames.
+        output_images (list[np.array]): The output frames.
+
+    Returns:
+        list[np.array]: The final frames.
+    """
+    if isinstance(output_images[0], list):
+        output_images = sum(output_images, [])
+    return input_images + output_images
```
ckkelvinchan, line 45: This is obsolete.

mmedit/apis/video_interpolation_inference.py
```diff
@@ -0,0 +1,134 @@
...
+        start_idx (int): The index corresponds to the first frame in the
+            sequence. Default: 0.
+        end_idx (int | None): The index corresponds to the last interpolated
+            frame in the sequence. If it is None, interpolate to the last
+            frame of video or sequence. Default: None.
+        batch_size (int): Batch size. Default: 4.
+
+    Returns:
+        output (list[numpy.array]): The predicted interpolation result.
+            It is an image sequence: [ori, pred, ori, pred, ori, ...]
```
ckkelvinchan, line 67: Not necessarily `[ori, pred, ori, pred, ori, ...]`

ckkelvinchan (approved):

PR #691, opened 2022-01-01
[Feature] CAIN model

Yshuo-Li:


mmedit/models/video_interpolater/cain.py
```diff
@@ -0,0 +1,109 @@
...
+        # generator
+        pred = self.generator(inputs, padding_flag=True)
+
+        # loss
+        losses = dict()
+        log_vars = dict()
+        losses['loss_pix'] = self.pixel_loss(pred, target)
+
+        # parse loss
+        loss, log_vars = self.parse_losses(losses)
```
ckkelvinchan, line 49: In `BasicInterpolater`, `self.parse_losses` is not done in this function. Is there any reason we do not follow `BasicInterpolater`?

mmedit/models/video_interpolater/cain.py
```diff
@@ -0,0 +1,109 @@
...
+        Returns:
+            dict: Output results, which contain either key(s)
+                1. 'eval_result'.
+                2. 'inputs', 'pred'.
+                3. 'inputs', 'pred', and 'target'.
+        """
+
+        # generator
+        with torch.no_grad():
+            pred = self.generator(inputs, padding_flag=True)
```
ckkelvinchan, line 93: So `padding_flag` is always True?

Yshuo-Li: No, `padding_flag` is False in training (in the original repo)

mmedit/models/video_interpolater/cain.py
Yshuo-Li: The subclasses should overwrite the function `forward_train`, `forward_test`, and `train_step`.

ckkelvinchan: `train_step` is not modified in this class. It seems that `self.parse_losses` is called twice. 

If the functionalities are identical to the parent class, I think we can directly inherit from it.

mmedit/models/video_interpolater/cain.py
ckkelvinchan: But it is also set to True in training? (line 41)

mmedit/models/video_interpolater/cain.py
```diff
@@ -0,0 +1,109 @@
...
+            dict: Returned output, which includes:
+                log_vars, num_samples, results (inputs, target and pred).
+
+        """
+        # data
+        inputs = data_batch['inputs']
+        target = data_batch['target']
+
+        # generator
+        pred = self.generator(inputs, padding_flag=False)
```
ckkelvinchan, line 41: @Yshuo-Li I think the only difference between this function and that in `BasicInterpolator` is this line, where an additional argument `padding_flag=False` is required? If yes, I think you can simply copy the whole function to here, and then modify this line. 

Another way is to set a default value in the backbone. If the default value is False, then this function can be inherited from `BasicInterpolator`. If the default is True. Then `forward_test` can be inherited.

ckkelvinchan (approved):

ckkelvinchan:
ImportError: cannot import name 'PILLOW_VERSION' from 'PIL' (/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/PIL/__init__.py)

It seems that the previous errors occur. Maybe try pulling from master and see if the problem persists

ckkelvinchan (approved):

PR #695, opened 2022-01-04
[Feature] Add BaseVFIDataset

Yshuo-Li:


ckkelvinchan (commented):
Some lines are not covered by unittests. 

Moreover, please check if a dataset contains multiple videos with different number of frames, how the PSNR is calculated. In video SR, we usually compute the average PSNR of each video, and then take the average of all videos, as shown [here](https://github.com/open-mmlab/mmediting/blob/master/mmedit/datasets/sr_folder_video_dataset.py#L170).

Yshuo-Li:
> Some lines are not covered by unittests.
> 
> Moreover, please check if a dataset contains multiple videos with different number of frames, how the PSNR is calculated. In video SR, we usually compute the average PSNR of each video, and then take the average of all videos, as shown [here](https://github.com/open-mmlab/mmediting/blob/master/mmedit/datasets/sr_folder_video_dataset.py#L170).

If a dataset contains multiple videos with different numbers of frames, it should inherit the basic class and modify this function.

ckkelvinchan:
> > Some lines are not covered by unittests.
> > Moreover, please check if a dataset contains multiple videos with different number of frames, how the PSNR is calculated. In video SR, we usually compute the average PSNR of each video, and then take the average of all videos, as shown [here](https://github.com/open-mmlab/mmediting/blob/master/mmedit/datasets/sr_folder_video_dataset.py#L170).
> 
> If a dataset contains multiple videos with different numbers of frames, it should inherit the basic class and modify this function.

So this dataset is for `N frames -> 1 frame` case?

Yshuo-Li:
> > > Some lines are not covered by unittests.
> > > Moreover, please check if a dataset contains multiple videos with different number of frames, how the PSNR is calculated. In video SR, we usually compute the average PSNR of each video, and then take the average of all videos, as shown [here](https://github.com/open-mmlab/mmediting/blob/master/mmedit/datasets/sr_folder_video_dataset.py#L170).
> > 
> > 
> > If a dataset contains multiple videos with different numbers of frames, it should inherit the basic class and modify this function.
> 
> So this dataset is for `N frames -> 1 frame` case?

Yes, 2 frames -> 1 frame

ckkelvinchan (approved):

PR #701, opened 2022-01-11
Deprecate the support for "python setup.py test"

ckkelvinchan:
Reference: https://github.com/open-mmlab/mmcv/pull/1637


Yshuo-Li (approved):

wangruohui:
We can also remove `requirements/tests.txt` actually.

PR #719, opened 2022-01-27
[Feature] Configuration files and checkpoints for EDVR-L on REDS

ckkelvinchan:
## Motivation
To support EDVR-L on REDS. 

## Modification
1. Add two configuration files [here](https://github.com/open-mmlab/mmediting/tree/master/configs/restorers/edvr)
2. Update README

## Note
The uploaded configuration file is 0.04 dB lower than the official model, which is trained with a lot of finetuning and is not able to be reproduced for now.


configs/restorers/edvr/README.md
```diff
@@ -37,3 +37,15 @@ The metrics are `PSNR / SSIM` .
   year      = {2019},
 }
 ```
+
+## Results and models
```
Yshuo-Li, line 41: Double `Results and models`.
Please see `https://github.com/open-mmlab/mmediting/blob/master/configs/restorers/edsr/README.md` for the readme format

Yshuo-Li (approved):

PR #720, opened 2022-01-27
[Feature] Support CPU training

ckkelvinchan:
Reference: https://github.com/open-mmlab/mmdetection/pull/7016


wangruohui:
Should we also modify somewhere in the documents ?

ckkelvinchan:
> Should we also modify somewhere in the documents ?

Let me check

wangruohui (approved):

PR #725, opened 2022-02-08
Rearrange requirements.txt

wangruohui:
Update of Windows and Linux CI are split to #732 and #734.
This PR deals with requirements.txt only. 

1. merge *av* and *facexlib* into `optional.txt` and use `pip install -e .[all]` to install all dependencies. 
1. pass test on Python 3.6.7, which is default of ubuntu18.04


Yshuo-Li (approved):

wangruohui:
I met same problem as described here
https://github.com/facebook/jest/issues/9413

wangruohui:
After amount of testing, on windows, when running multiple commands, only when the last one exit error, the test will fail. So I split unit test and coverage

wangruohui:
Reopened in to #732 and #734

PR #758, opened 2022-02-24
[Docs] Add tutorial of dataset

Yshuo-Li:


wangruohui:
1. An item on index.rst is required to show it on the index page.
1. We need to describe, when inheriting from base dataset, which method need to be overwritten ( load annotaion here) and what result is desired to be returned.

wangruohui:
Another suggestion: 
As several existing dataset classes are referenced, we can try cross reference functionality.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-class

wangruohui (approved):

wangruohui (approved):

PR #783, opened 2022-03-11
add 'DIM online merge config demo'

CCODING04:
I checked the DIM demo config, and found that there is no "online merge" demo. So I create it for convenience. 
MMEditing have already realized the code, but not shown us how to use it.  Maybe dataset doc file could be detailed.


wangruohui:
Thanks you for your contribution, we will arrange to review these codes.

CCODING04:
> 

Unit test or just a simple use case?

wangruohui:
@CCODING04 

A simple case is OK. 
Have you trained with this configuration? Could you please upload the training log and the file of pre-trained weight?

CCODING04:
@wangruohui 
I'm training it. However, it seems like merge online is not a good choice cuz the curve of loss and validation metrics got violate vibration.

CCODING04:
@wangruohui 
Actually, I did previously merge and got a good result.
Is there any paper or articles talk about which and when to take composition of data is better?

CCODING04:
DIM_cmp_MSE
DIM_cmp_SAD
@wangruohui 
I compared the online and offine results and plot it.

zengyh1900:
Please check this. @Z-Fran

Z-Fran:
Hi @CCODING04 , we plan to merge your code to MMEditing 2.0. So you need to do two jobs as following:
1. modify file name and directory to `configs/dim/dim_stage1-v16_1xb1-1000k_comp1k_online-merge.py`
2. If there is no difference between [dim_stage1-v16_1xb1-1000k_comp1k.py](https://github.com/open-mmlab/mmediting/blob/dev-1.x/configs/dim/dim_stage1-v16_1xb1-1000k_comp1k.py) and your code except `train pipeline`, we suggest you can inherit the same codes, like `
_base_ = ['./dim_stage1-v16_1xb1-1000k_comp1k.py']` in [dim_stage2-v16-pln_1xb1-1000k_comp1k.py](https://github.com/open-mmlab/mmediting/blob/dev-1.x/configs/dim/dim_stage2-v16-pln_1xb1-1000k_comp1k.py). Then you can delete the same parts.

And you can give us authority to commit codes to your respository to help you. Thanks again.

zengyh1900 (approved):

PR #792, opened 2022-03-16
[Enhance] Put LQ into CPU for restoration_video_demo

ckkelvinchan:
## Motivation
Instead of loading all LQ images to GPU, which causes CUDA OOM, we first put them in CPU, and then extract to GPU when we use them.

## Modifiication
- Modify `mmedit/apis/restoration_video_inference.py`

### DIFF
--- a/mmedit/apis/restoration_video_inference.py
+++ b/mmedit/apis/restoration_video_inference.py
@@ -1,13 +1,10 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import glob
-import os.path as osp
-import re
-from functools import reduce
+import os
 
 import mmcv
 import numpy as np
 import torch
-from mmcv.parallel import collate, scatter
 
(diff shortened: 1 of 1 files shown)


mmedit/apis/restoration_video_inference.py
```diff
@@ -91,10 +88,9 @@ def restoration_video_inference(model,
         test_pipeline[0]['filename_tmpl'] = filename_tmpl
 
         # prepare data
-        sequence_length = len(glob.glob(osp.join(img_dir, '*')))
-        img_dir_split = re.split(r'[\\/]', img_dir)
-        key = img_dir_split[-1]
-        lq_folder = reduce(osp.join, img_dir_split[:-1])
```
wangruohui, line 97: These lines are useful to handle path separator `\` on Windows.

wangruohui (changes requested):

wangruohui (approved):

PR #798, opened 2022-03-17
Use full name of config as 'Name' in metafile

Yshuo-Li:


wangruohui (approved):

wangruohui:
This is modified to make mim worked correctly.

PR #799, opened 2022-03-17
[Fix] `flip_ratio` in `RandomAffine` pipeline is inverted

wdmwhh:
original: `flip_ratio=1.` means always no flip and `flip_ratio=0.` means always flip
after:      `flip_ratio=1.` means always flip and `flip_ratio=0.` means always no flip

### DIFF
--- a/mmedit/datasets/pipelines/augmentation.py
+++ b/mmedit/datasets/pipelines/augmentation.py
@@ -480,7 +480,6 @@ class RandomAffine:
 
         angle = math.radians(angle)
         shear = math.radians(shear)
-        # print(flip)
         scale_x = 1.0 / scale[0] * flip[0]
         scale_y = 1.0 / scale[1] * flip[1]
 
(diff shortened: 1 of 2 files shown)


wangruohui:
Because `flip` is used as a multiplier in line 479 and 480, so -1 stands for flip and 1 stands for no flip. Thus `flip` should be an 'inverse' flag as the result of the comparison. 

Is it correct?

wdmwhh:
> Because `flip` is used as a multiplier in line 479 and 480, so -1 stands for flip and 1 stands for no flip. Thus `flip` should be an 'inverse' flag as the result of the comparison.
> 
> Is it correct?

Yes, it is.

wangruohui:
It seems there is another error. When flipping, the center is shifted. 
I made some modifications and added some unit tests, you can have a test to verify.

wdmwhh:
> It seems there is another error. When flipping, the center is shifted. I made some modifications and added some unit tests, you can have a test to verify. I am not sure whether it is related to version of opencv
> 
> @wdmwhh, if the modification works, please let me know and I will clean the code and merge this PR.

Yes, it makes sense. Thanks for your kindness.

wangruohui (approved):

PR #800, opened 2022-03-17
[Docs] Add brief installation steps in README(_zh-CN).md

wangruohui:


README.md
```diff
@@ -82,7 +82,34 @@ Please refer to [changelog.md](docs/en/changelog.md) for details and release his
 
 ## Installation
 
-Please refer to [install.rst](docs/en/install.rst) for installation.
+MMEditing depends on [PyTorch](https://pytorch.org/) and [MMCV](https://github.com/open-mmlab/mmcv).
+Below are quick steps for installation.
+
+**Step 1.**
+Install PyTorch following [official instructions](https://pytorch.org/get-started/locally/), e.g.
```
ZwwWayne, line 89: I suggest using command:
```
conda create -n open-mmlab python=3.7 -y
conda activate open-mmlab
conda install cudatoolkit=11.0 pytorch=1.9 -c pytorch -y
pip install openmim
mim install mmcv-full
git clone https://github.com/open-mmlab/mmediting.git
cd mmediting && pip install -e .
```

wangruohui: I think three pieces of code are better. 
- Github provides functions to copy by a single click, so coping three pieces does not require more effort. 
- Multiple commands are buggier when they are interrupted by errors. Command followed will still execute to cause more errors. For example if `conda` commands failed, the following `pip install` will install to another environment (probably base or the OS's python environment) and may even break that environment.
- Multiple sections separated by functionalities are clear to show and we can also add some descriptive words.

If this is considered too long to show in README we can move it to docs. 
And if one prefer `conda install pytorch`  instead of pip, this PR #816 changed it.

wangruohui:
#816 further modified

PR #808, opened 2022-03-21
[Fix] use hyphen for command line args in demo & tools 

nijkah:
closes https://github.com/open-mmlab/mmediting/issues/804

### DIFF
--- a/demo/generation_demo.py
+++ b/demo/generation_demo.py
@@ -11,10 +11,10 @@ def parse_args():
     parser = argparse.ArgumentParser(description='Generation demo')
     parser.add_argument('config', help='test config file path')
     parser.add_argument('checkpoint', help='checkpoint file')
-    parser.add_argument('img_path', help='path to input image file')
-    parser.add_argument('save_path', help='path to save generation result')
+    parser.add_argument('img-path', help='path to input image file')
+    parser.add_argument('save-path', help='path to save generation result')
     parser.add_argument(
-        '--unpaired_path', default=None, help='path to unpaired image file')
+        '--unpaired-path', default=None, help='path to unpaired image file')
     parser.add_argument(
         '--imshow', action='store_true', help='whether show image with opencv')
     parser.add_argument('--device', type=int, default=0, help='CUDA device id')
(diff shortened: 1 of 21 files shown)


Yshuo-Li:
Good job.

Maybe markdown files such as `docs/en/demo.md`.

Yshuo-Li:
We can find all `--xx_x` by the following command

`grep -rn -E -- '--[[:alnum:]]+_' docs`

nijkah:
> We can find all `--xx_x` by the following command
> 
> `grep -rn -E -- '--[[:alnum:]]+_' docs`

I ran 
```
grep -e "--\w*_" -rn demo/*.py 
grep -e "--\w*_" -rn tools
grep -e "--\w*_" -rn docs
```

and added support to previous arguments.
+ Lint

wangruohui:
@Yshuo-Li 

Would the following implementation be better? We can use only hyphens in `argparse`, and add an additional function to check and modify underscore `_` in-place in `sys.argv`. By doing so, the code would be kept clean.

We can put this function to `mmedit/utils/cli.py` and import them in demo scripts. 


save to `arg.py` and test with the command line.

```python
import warnings
import sys
import re


def parse_args():
    parser = argparse.ArgumentParser(description="Generation demo")
    parser.add_argument("--a-b", help="test config file path")
    parser.add_argument("--c-d-e", help="test config file path")
    parser.add_argument("--f", help="test config file path")
    args = parser.parse_args()
    return args


def main():
    print(sys.argv)
    modify_args()
    print(sys.argv)
    args = parse_args()
    print(args)


def modify_args():
    for i, v in enumerate(sys.argv):
        if i == 0:
            assert v.endswith(".py")
        elif re.match(r"--\w+_.*", v):
            new_arg = v.replace("_", "-")
            warnings.warn(
                f"command line argument {v} is deprecated, please use {new_arg} instead.",
                category=DeprecationWarning,
            )
            sys.argv[i] = new_arg


if __name__ == "__main__":
    main()
```

nijkah:
@wangruohui 
I updated it.
Note that `local_rank` in tools/train.py and tools/test.py is not modified because it is from pytorch's convention.

wangruohui:
@nijkah Thank you very much!

wangruohui (approved):

nijkah:
Should I add a test code for `modify_args`?

Yshuo-Li (approved):

Yshuo-Li:
> Should I add a test code for `modify_args`?

Yeah, that will be better.

Yshuo-Li (approved):

PR #815, opened 2022-03-23
[Feature] Support seed in DistributedSampler and sync seed across ranks

Yshuo-Li:


mmedit/datasets/samplers/distributed_sampler.py
```diff
@@ -40,6 +51,10 @@ def __iter__(self):
         # deterministically shuffle based on epoch
         if self.shuffle:
             g = torch.Generator()
+            # When :attr:`shuffle=True`, this ensures all replicas
+            # use a different random ordering for each epoch.
+            # Otherwise, the next iteration of this sampler will
+            # yield the same ordering.
             g.manual_seed(self.epoch)
```
LeoXing1996, line 58: According to  https://github.com/open-mmlab/mmdetection/blob/eb86b6538f2f52198c34860b24b517f9e78ae13d/mmdet/datasets/samplers/distributed_sampler.py#L37, here should be changed to `self.epoch + self.seed`

wangruohui (approved):

PR #817, opened 2022-03-24
[Improve] Update essential requirements

wangruohui:
- remove yapf and skimage
- add numpy and pillow
- add torch and torchvision


requirements/runtime.txt
```diff
@@ -1,12 +1,10 @@
...
 opencv-python<=4.5.4.60
 # MMCV depends opencv-python instead of headless, thus we install opencv-python
 # Due to a binary compatibility bug, we limit its version before 4.5.5
 # https://github.com/opencv/opencv-python/issues/602
 # https://github.com/opencv/opencv/issues/21366
 # It seems to be fixed in https://github.com/opencv/opencv/pull/21382
 # But our test is negative.
-scikit-image
 tensorboard
-yapf
```
Yshuo-Li, line 12: Why delete these two lines?

wangruohui: it seems this codebase no longer rely on scikit-image, I search `skimage`, and there is no code importing this, also CI passes. 
yapf is only a code formatter, it is not a runtime dependency.

Further, I think we need to add numpy and pillow. They are now installed as a dependency of torchvison, but we'd better to add them in requirements explicitly. I will append them and rename this PR.

PR #822, opened 2022-03-29
[Fix] Use underline for required aruments in python files.

Yshuo-Li:


wangruohui:
Fix bug introduced by #808

wangruohui (approved):

PR #857, opened 2022-04-21
[Feature] Support use_cache and backend in `LoadImageFromFileList`

Yshuo-Li:


wangruohui:
Is this used for speeding up loading or decoding?

Yshuo-Li:
> Is this used for speeding up loading or decoding?

Yes, it is.

Beside, LoadImageFromFileList should support pillow backend.

mmedit/datasets/pipelines/loading.py
```diff
@@ -160,10 +163,26 @@ def __call__(self, results):
         if self.save_original_img:
             ori_imgs = []
         for filepath in filepaths:
-            img_bytes = self.file_client.get(filepath)
-            img = mmcv.imfrombytes(
-                img_bytes, flag=self.flag,
-                channel_order=self.channel_order)  # HWC
+            if self.use_cache:
+                if self.cache is None:
+                    self.cache = dict()
```
wangruohui, line 168: this line should be done in __init__ instead of in a for loop, we don't need to do this every iteration. This will slow the program a little bit.

wangruohui (approved):

PR #859, opened 2022-04-22
[Feature] Update ColorJitter

Yshuo-Li:


mmedit/datasets/pipelines/augmentation.py
```diff
@@ -626,30 +627,52 @@ class ColorJitter:
...
+                ]
+            else:
+                results[k] = self._color_jitter(results[k], this_seed)
         return results
 
     def __repr__(self):
         repr_str = self.__class__.__name__
-        repr_str += (f'(keys={self.keys}, to_rgb={self.to_rgb})')
+        repr_str += (
+            f'(keys={self.keys}, channel_order={self.channel_order}, ')
```
wangruohui, line 675: Directly `repr_str = f'{self.__class__.__name__}(keys={self.keys}, channel_order={self.channel_order})'`

tests/test_data/test_pipelines/test_augmentation.py
```diff
@@ -364,18 +364,40 @@ def test_color_jitter(self):
...
+            saturation=0.5,
+            hue=0.5)
         color_jitter_results = color_jitter(results)
         assert self.check_keys_contain(color_jitter_results.keys(),
                                        target_keys)
         assert color_jitter_results['gt'].shape == self.img_gt.shape
 
         assert repr(color_jitter) == color_jitter.__class__.__name__ + (
-            f"(keys=['gt'], to_rgb=False)")
+            "(keys=['gt', 'lq'], channel_order=bgr)")
```
wangruohui, line 391: `channel_order=bgr` should be `channel_order='bgr'`. 
Besides, its better to hardcode here, 
```
assert repr(color_jitter) == "ColorJitter(keys=['gt', 'lq'], channel_order='bgr')"
```

tests/test_data/test_pipelines/test_augmentation.py
```diff
@@ -364,18 +364,40 @@ def test_color_jitter(self):
...
+        target_keys = ['gt', 'lq']
 
         color_jitter = ColorJitter(
-            keys=['gt'], brightness=0.5, contrast=0.5, saturation=0.5, hue=0.5)
+            keys=['gt', 'lq'],
+            brightness=0.5,
+            contrast=0.5,
+            saturation=0.5,
+            hue=0.5)
+        color_jitter_results = color_jitter(results)
```
wangruohui, line 377: This result is overwritten in line 385 before the assertion. We need to add some assert for this result.

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -626,30 +627,51 @@ class ColorJitter:
 
     Args:
         keys (list[str]): The images to be resized.
-        to_rgb (bool): Whether to convert channels from BGR to RGB.
-            Default: False.
+        channel_order (str): Order of channel, candidates are 'bgr' and 'rgb'.
+            Default: 'rgb'.
```
wangruohui, line 631: The docstring should notify users that` brightness`, `hue` etc. are also acceptable arguments. if `kwargs` is directly passed to `torchvision.trasnforms.ColorJitter`, we can simply tell users to follow args of `torchvision.transforms.ColorJitter`. And for the same reason, `__repr__` usually contain full kwargs, so the expression returned by `__repr__` constructs the same transformation object.

mmedit/datasets/pipelines/augmentation.py
```diff
@@ -626,30 +627,51 @@ class ColorJitter:
...
+        image = np.asarray(image)
+
+        if self.channel_order.lower() == 'bgr':
+            image = image[..., ::-1]
+
+        return image
+
     def __call__(self, results):
+
+        this_seed = random.randint(0, 2**32)
```
wangruohui, line 661: Why do we need to again random a seed? They are originally random.

tests/test_data/test_pipelines/test_augmentation.py
```diff
@@ -364,18 +364,40 @@ def test_color_jitter(self):
...
+            "(keys=['gt', 'lq'], channel_order=bgr)")
+        with pytest.raises(AssertionError):
+            color_jitter = ColorJitter(
+                keys=['gt', 'lq'],
+                channel_order='bgr',
+                to_rgb=True,
+                brightness=0.5,
+                contrast=0.5,
+                saturation=0.5,
+                hue=0.5)
```
wangruohui, line 401: We'd better add some assertion to verify that the resulted image array is **different** from the input ones, to show that this transformation is at least functional.

mmedit/datasets/pipelines/augmentation.py
Yshuo-Li: To ensure all frames are processed by ColorJitter with the same parameters.

wangruohui (approved):

PR #860, opened 2022-04-22
[Feature] Support ReduceLrUpdaterHook

Yshuo-Li:


mmedit/core/scheduler/lr_updater.py
```diff
@@ -50,3 +50,246 @@ def get_lr(self, runner, base_lr):
...
+                 **kwargs):
+
+        self.val_metric = val_metric
+
+        if mode not in ['min', 'max']:
+            raise ValueError(
+                'mode must be one of "min" or "max", instead got {mode}')
+        self.mode = mode
+
+        if factor >= 1.0:
```
wangruohui, line 121: maybe `if factor >= 1.0 or factor <0` is better?

mmedit/core/scheduler/lr_updater.py
```diff
@@ -50,3 +50,246 @@ def get_lr(self, runner, base_lr):
...
+                self.num_bad_epochs = 0
+            else:
+                self.num_bad_epochs += 1
+
+            if self.in_cooldown:
+                self.cooldown_counter -= 1
+                self.num_bad_epochs = 0
+            print(f'train_iter--current {current:.6f} best {self.best:.6f}, '
+                  f'num_bad_epochs {self.num_bad_epochs}, '
+                  f'cooldown {self.in_cooldown} {self.cooldown_counter}')
```
wangruohui, line 252: If `val_metric` is specified, we check its value after training iteration based on **training loss**, else, we check it after validation iteration based on some **validation metrics**. Is it right?

We should make this clear in the docstring, otherwise, it's confusing, because it may engage in different hooks.

tests/test_runtime/test_lr_updater.py
```diff
@@ -0,0 +1,110 @@
...
+def test_linear_lr_updater_hook():
+    fake_runner = FakeRunner()
+
+    lr_updater = LinearLrUpdaterHook(by_epoch=False)
+    lr_updater.get_lr(fake_runner, 1)
+
+    lr_updater = LinearLrUpdaterHook(by_epoch=True)
+    lr_updater.get_lr(fake_runner, 1)
+    lr_updater.start = 10
+    lr_updater.get_lr(fake_runner, 1)
```
wangruohui, line 34: We need to at least assert something in the test, e.g. if the learning rate reduces to a desired value.
Same for `test_reduce_lr_updater_hook`

wangruohui: Is this another feature irrelevant to reducing on the plateau?

Yshuo-Li: > Is this another feature irrelevant to reducing on the plateau?

Yes, it is another feature irrelevant to reducing on the plateau, which is in lack of unit test

wangruohui:
We still need to assert the `lr` get from `lr_updater.get_lr(fake_runner, 1)` is expected. They are missing currently.

wangruohui (approved):

PR #861, opened 2022-04-24
[Feature] Support `after_val_epoch` in `IterBaseRunner`

Yshuo-Li:


mmedit/core/evaluation/eval_hooks.py
```diff
@@ -59,6 +59,8 @@ def evaluate(self, runner, results):
         for name, val in eval_res.items():
             runner.log_buffer.output[name] = val
         runner.log_buffer.ready = True
+        # call `after_val_epoch` after evaluation.
+        runner.call_hook('after_val_epoch')
```
wangruohui, line 63: explain more. 

This is a hack. Because epoch does not naturally exist In `iterbasedrunner`, thus we consider the **end** of an **evluation** as the **end of an epoch**. With this hack , we can support epoch based hooks.

wangruohui (approved):

wangruohui (approved):

PR #862, opened 2022-04-24
[Fix] Modify double `lq_path` to `gt_path` in `test_pipeline`

ryanxingql:
Some `meta_keys` in config files might should be `['lq_path', 'gt_path']` instead of `['lq_path', 'lq_path']`.

PS Two ipynb files [1](https://github.com/open-mmlab/mmediting/blob/master/demo/restorer_basic_tutorial.ipynb) and [2](https://github.com/open-mmlab/mmediting/blob/master/demo/restorer_basic_tutorial_zh-CN.ipynb) should be re-run to fix the typos. I haven't done that yet.

### DIFF
--- a/configs/restorers/esrgan/esrgan_psnr_x4c64b23g32_g1_1000k_div2k.py
+++ b/configs/restorers/esrgan/esrgan_psnr_x4c64b23g32_g1_1000k_div2k.py
@@ -65,7 +65,7 @@ test_pipeline = [
         mean=[0, 0, 0],
         std=[1, 1, 1],
         to_rgb=True),
-    dict(type='Collect', keys=['lq', 'gt'], meta_keys=['lq_path', 'lq_path']),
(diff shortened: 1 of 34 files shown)


wangruohui:
I think this is not a typo. In test pipeline, we don't actually have the `gt_path`.

Yshuo-Li:
> I think this is not a typo. In `test_pipeline`, we don't actually have the `gt_path`.

This is a typo.

1. If there is no `gt_path`, the test pipeline cannot load `gt` by `LoadImageFromFile`.
2. In demo scripts, if the configure file doesn't contain `demo_pipeline`, `test_pipeline` will be used for inference. However, demo scripts will remove `gt` and `gt_path` from the pipeline.

Besides, the reason why this typo doesn't raise any error is that `gt_path` is not used in the follow-up processing.

Yshuo-Li (approved):

PR #888, opened 2022-05-17
[Feature] use mdformat

wangruohui:


README.md
```diff
@@ -76,13 +75,13 @@ hope MMEditing could provide better experience.
 
 ## News
 
-- [2022-04-01] v0.14.0 was released.
+- \[2022-04-01\] v0.14.0 was released.
```
Yshuo-Li, line 78: No change in the preview.

Yshuo-Li (approved):

README.md
wangruohui: This is as expected, the formatter adds backslash `\` to `[]` to make sure these brackets should not be parsed as links.

PR #965, opened 2022-07-20
Support c=1 images in `FramesToTensor`

ryanxingql:
Hi @ckkelvinchan 
When I used `FramesToTensor` for my gray-scale images, I met a bug. And I solve it in this PR.

In specific, I want to train and test my model over the Y data.
My config file is in [here](https://github.com/ryanxingql/powervqe/blob/a3b63b9311660226146a26902bbb94aa48fe22c5/mmediting/configs/restorers/stdf/ldv_v2_y_4gpus.py#L55).
Considering the original data is in the RGB format, I add the `convert_to='y'` option in both the `train_pipeline` and `test_pipeline`.
The ground truth tensor should be with the shape of `(channels=1, height, weight)` for calculating losses or for evaluation.
However, due to [this line](https://github.com/open-mmlab/mmediting/blob/48ca8e1707e203445c296f363c5a4c9fc6b09bf3/mmedit/datasets/pipelines/formating.py#L137), the tensor is squeezed into `(height, weight)`.
So I specify the `dim=0` option in accord with the `if results[key].size(0) == 1` check.

BTW I add `FramesToTensor` to the` __init__` file.


ryanxingql:
@wangruohui Hi ruohui, can you help me review this PR? Thank you.

PR #975, opened 2022-07-25
[Fix] fix model registry and docs in demo

Z-Fran:
1. fix model registry in init_model
2. use checkpoint link in all demo
3. fix config file path

### DIFF
--- a/mmedit/apis/matting_inference.py
+++ b/mmedit/apis/matting_inference.py
@@ -6,6 +6,7 @@ from mmcv.runner import load_checkpoint
 from mmengine.dataset import Compose
 
 from mmedit.models import build_model
+from mmedit.registry import register_all_modules
 
 
 def init_model(config, checkpoint=None, device='cuda:0'):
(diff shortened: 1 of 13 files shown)


mmedit/apis/matting_inference.py
```diff
@@ -21,12 +22,14 @@ def init_model(config, checkpoint=None, device='cuda:0'):
...
     if isinstance(config, str):
         config = mmcv.Config.fromfile(config)
     elif not isinstance(config, mmcv.Config):
         raise TypeError('config must be a filename or Config object, '
                         f'but got {type(config)}')
-    config.model.pretrained = None
+    if 'pretrained' in config.model:
+        config.model.pretrained = None
     config.test_cfg.metrics = None
     model = build_model(config.model, test_cfg=config.test_cfg)
```
wangruohui, line 34: We will finally deprecate `build_model` APIs. In the new design, we will use the following instead.

```
from mmedit.registry import MODELS
MODELS.build(cfg)
```

mmedit/apis/matting_inference.py
```diff
@@ -21,12 +22,14 @@ def init_model(config, checkpoint=None, device='cuda:0'):
     Returns:
         nn.Module: The constructed model.
     """
+    register_all_modules(init_default_scope=True)
     if isinstance(config, str):
         config = mmcv.Config.fromfile(config)
     elif not isinstance(config, mmcv.Config):
         raise TypeError('config must be a filename or Config object, '
                         f'but got {type(config)}')
-    config.model.pretrained = None
```
wangruohui, line 29: In the new design, pretrain is no longer an item in model. Instead, we will use `init_cfg`.

```
model = dict(
type='xxx'
init_cfg=dict(
type='Pretrain', 
weight='some_path'
)
```
like `https://github.com/open-mmlab/mmediting/blob/dev-1.x/configs/mattors/gca/baseline_r34_4x10_200k_comp1k.py`

Pre-train weight is useless and so we can find `init_cfg` recursively and delete them.

Z-Fran: done

mmedit/apis/matting_inference.py
Z-Fran: done

wangruohui (approved):

PR #1054, opened 2022-08-26
[Refactor] refactor data flow and engine library

zengyh1900:


plyfager (commented):
build_activation_layer, build_conv_layer, build_norm_layer is no longer used, maybe find and replace them with MODELS.build instead.(seems not necessary now)

LeoXing1996 (approved):

PR #1058, opened 2022-08-28
[Document] user_guides/train_test.md 

zengyh1900:


plyfager (commented):
The old version ``train_test`` has several mistakes for current version. Maybe a overall check is needed.

plyfager:
1. ``--save-path`` seems not supported. 
2. may check whether the url is validate and refers to 1.x version.
3. evaluation config example seems not correct.
4. --resume, not --resume-from
5. ``dist_params`` seems not correct.

PR #1059, opened 2022-08-28
[Documentation] useful tools: flops, model converters

zengyh1900:


docs/en/user_guides/useful_tools.md
```diff
@@ -1,19 +1,19 @@
-## Useful tools
+# Useful tools
 
 We provide lots of useful tools under `tools/` directory.
 
-### Get the FLOPs and params (experimental)
+## Get the FLOPs and params
```
plyfager, line 5: I'm not sure whether params equal to parameter amount. Maybe we can introduce what is ``FLOPs`` and ``params`` and why we need these information for starters.

docs/en/user_guides/useful_tools.md
```diff
@@ -32,30 +32,30 @@ Params: 1.52 M
...
 ```shell
-python tools/publish_model.py ${INPUT_FILENAME} ${OUTPUT_FILENAME}
+python tools/model_converters/publish_model.py ${INPUT_FILENAME} ${OUTPUT_FILENAME}
 ```
 
 E.g.,
 
 ```shell
-python tools/publish_model.py work_dirs/example_exp/latest.pth example_model_20200202.pth
+python tools/model_converters/publish_model.py work_dirs/example_exp/latest.pth example_model_20200202.pth
```
plyfager, line 48: maybe we should give time tag by script. Example: https://github.com/open-mmlab/mmgeneration/blob/master/tools/publish_model.py#L29.

plyfager (commented):
The table in ``List of supported models exportable to ONNX`` has a column note, but seems the  note contain nothing, can we remove this column.

docs/en/user_guides/useful_tools.md
zengyh1900: > maybe we should give time tag by script. Example: https://github.com/open-mmlab/mmgeneration/blob/master/tools/publish_model.py#L29.

to be done in the other PR. 
In this PR, we only discuss docs.

PR #1083, opened 2022-09-05
[Fix] fix multi_loops UT and clean debug code

Z-Fran:


mmedit/apis/video_interpolation_inference.py
```diff
@@ -192,7 +192,6 @@ def video_interpolation_inference(model,
         if start_index + lenth_per_step >= end_idx:
             break
 
-    print()
```
zengyh1900, line 195: I think it is used for creating a new line, instead of temporal debug. We may need it here.

Z-Fran: done

zengyh1900 (approved):

PR #1089, opened 2022-09-07
Review by Wu Bohuai

VongolaWu:
It is a little strange to capitalize both E and M for Easy-Modular.


zengyh1900:
This has been resolved after merging the docs of MMGeneration into MMEditing.

PR #1093, opened 2022-09-08
Review by Bohuai

VongolaWu:
The original gpu installation command will cause the cpu version installed for no reason. I don't think `cd ~` is needed, because they may not want to go back to home folder. Change some sentances.


zengyh1900:
The related issue been resolved in https://github.com/open-mmlab/mmediting/pull/1094

PR #1332, opened 2022-10-18
[Bug] Correct RDN number of channels

ryanxingql:
## Motivation

The RDN implementation will cause error when `channel_growth != mid_channels`.

## Modification

We have fixed the bug of RDN in V0.15 (PRs https://github.com/open-mmlab/mmediting/pull/1292 and https://github.com/open-mmlab/mmediting/pull/1311). Also, PR https://github.com/open-mmlab/mmediting/pull/1326 updates the docstring for RDN. Now we adopt these changes into V1.x.

We also add the `channel_growth` option into the cfg of the RDN test file.

@LeoXing1996


zengyh1900 (approved):
Good job!

PR #1484, opened 2022-11-29
[Fix] Wrong path results in pre-commit error in python3.9

liangzelong:
## Motivation

In mmediting master branch "update_model_index.py" locates in "mmediting/.dev_scripts/github" while in 1.x and dev-1.x branch it locates in "mmediting/.dev_scripts". But the code remains. Before python 3.9, attrbute  "\_\_file\_\_"  uses a relative path. Pre-commit runs well.  But in python 3.9 it uses an absolute path resulting in errors in pre-commit.

## Modification

change the wrong path and ensure pre-commit runs well in python 3.9.


LeoXing1996:
LGTM!

LeoXing1996 (approved):

PR #1491, opened 2022-12-01
[Refactor] use mmengine.basemodule instead of nn.module

Z-Fran:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

[Refactor] use mmengine.basemodule instead of nn.module

## Modification

Use mmengine.basemodule instead of nn.module

- [x] 1. basicvsr and basicvsr++
- [x] 2. aot-gan
- [x] 3. cain
- [x] 4. deepfillv1_
- [x] 5. deepfillv2
- [x] 6. dim
- [x] 7. edsr
- [x] 8. edvr
- [x] 9. gl
- [x] 10. iconvsr
- [x] 11. indexnet
- [x] 12. liif-
- [x] 13. pconv
- [x] 14. rdn
- [x] 15. realbasicvsr
- [x] 16. realesrnet_
- [x] 17. realesrgan
- [x] 18. srcnn
- [x] 19. msrresnet_
- [x] 20. srgan
- [x] 21. tdan

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

- [x] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmediting/blob/master/.github/CONTRIBUTING.md) to create this PR.
- [x] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmediting/blob/master/.github/CONTRIBUTING.md) are used to fix the potential lint issues.
- [x] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [x] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [x] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [x] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects.
- [x] CLA has been signed and all committers have signed the CLA in this PR.


plyfager (approved):

mmedit/models/editors/deepfillv1/deepfill_disc.py
```diff
@@ -47,22 +46,11 @@ def forward(self, x):
 
         return global_pred, local_pred
 
-    def init_weights(self, pretrained=None):
-        """Init weights for models.
+    def init_weights(self):
+        """Init weights for models."""
```
LeoXing1996, line 50: I think we should add the following code here:
```python
        if self.init_cfg is None and self.init_cfg['type'] == 'Pretrained':
            super().init_weights()
            return
```
Otherwise, this model can never load the pre-trained model.

mmedit/models/editors/deepfillv2/two_stage_encoder_decoder.py
```diff
@@ -73,22 +72,11 @@ def forward(self, x):
         return stage1_res, stage2_res
 
     # TODO: study the effects of init functions
-    def init_weights(self, pretrained=None):
-        """Init weights for models.
+    def init_weights(self):
+        """Init weights for models."""
```
LeoXing1996, line 76: I think we should add the following code here:
```python
        if self.init_cfg is None and self.init_cfg['type'] == 'Pretrained':
            super().init_weights()
            return
```
Otherwise, this model can never load the pre-trained model.

mmedit/models/editors/global_local/gl_disc.py
```diff
@@ -47,22 +46,12 @@ def forward(self, x):
 
         return pred
 
-    def init_weights(self, pretrained=None):
-        """Init weights for models.
+    def init_weights(self):
+        """Init weights for models."""
```
LeoXing1996, line 50: I think we should add the following code here:
```python
        if self.init_cfg is None and self.init_cfg['type'] == 'Pretrained':
            super().init_weights()
            return
```
Otherwise, this model can never load the pre-trained model.

LeoXing1996 (approved):

PR #1637, opened 2023-02-10
[Fix] dynamic exportable pixel unshuffle

yhna940:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

These are changes to support pixel unshuffling for inputs of various sizes during onnx export. The warning is as follows:
```
/opt/conda/lib/python3.8/site-packages/mmedit/models/common/downsample.py:14: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if h % scale != 0 or w % scale != 0:
/opt/conda/lib/python3.8/site-packages/mmedit/models/common/downsample.py:18: TracerWarning: Converting a tensor to a Python integer might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  h = int(h / scale)
/opt/conda/lib/python3.8/site-packages/mmedit/models/common/downsample.py:19: TracerWarning: Converting a tensor to a Python integer might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  w = int(w / scale)
```

More detail could be found in here:
https://pytorch.org/docs/stable/onnx.html#avoid-numpy-and-built-in-python-types

## Modification

Please briefly describe what modification is made in this PR.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

- [ ] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmediting/blob/master/.github/CONTRIBUTING.md) to create this PR.
- [ ] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmediting/blob/master/.github/CONTRIBUTING.md) are used to fix the potential lint issues.
- [ ] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [ ] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [ ] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [ ] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects.
- [ ] CLA has been signed and all committers have signed the CLA in this PR.

### DIFF
--- a/mmedit/models/common/downsample.py
+++ b/mmedit/models/common/downsample.py
@@ -1,7 +1,4 @@
 # Copyright (c) OpenMMLab. All rights reserved.
-import torch
-
-
 def pixel_unshuffle(x, scale):
     """Down-sample by pixel unshuffle.
 
@@ -20,7 +17,6 @@ def pixel_unshuffle(x, scale):
             f'with shape: {x.shape}')
     h = h // scale
     w = w // scale
-    size = torch.Size([b, c, h, scale, w, scale])
-    x = x.view(size)
+    x = x.view([b, c, h, scale, w, scale])
     x = x.permute(0, 1, 3, 5, 2, 4)
     return x.reshape(b, -1, h, w)
--- a/mmedit/models/common/downsample.py
+++ b/mmedit/models/common/downsample.py
@@ -1,4 +1,7 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+import torch
+
+
 def pixel_unshuffle(x, scale):
     """Down-sample by pixel unshuffle.
 
@@ -15,8 +18,8 @@ def pixel_unshuffle(x, scale):
         raise AssertionError(
             f'Invalid scale ({scale}) of pixel unshuffle for tensor '
             f'with shape: {x.shape}')
-    h = int(h / scale)
-    w = int(w / scale)
+    h = torch.div(h, scale, rounding_mode='floor')
+    w = torch.div(w, scale, rounding_mode='floor')
     x = x.view(b, c, h, scale, w, scale)
     x = x.permute(0, 1, 3, 5, 2, 4)
     return x.reshape(b, -1, h, w)
--- a/mmedit/models/common/downsample.py
+++ b/mmedit/models/common/downsample.py
@@ -18,8 +18,9 @@ def pixel_unshuffle(x, scale):
         raise AssertionError(
             f'Invalid scale ({scale}) of pixel unshuffle for tensor '
             f'with shape: {x.shape}')
-    h = torch.div(h, scale, rounding_mode='floor')
-    w = torch.div(w, scale, rounding_mode='floor')
-    x = x.view(b, c, h, scale, w, scale)
+    h = h // scale
+    w = w // scale
+    size = torch.Size([b, c, h, scale, w, scale])
+    x = x.view(size)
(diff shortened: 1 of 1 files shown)


Z-Fran:
Please fix the CI error `FAILED tests/test_models/test_video_interpolator/test_cain.py::test_cain - TypeError: div() got an unexpected keyword argument 'rounding_mode'` @yhna940

yhna940:
> Please fix the CI error `FAILED tests/test_models/test_video_interpolator/test_cain.py::test_cain - TypeError: div() got an unexpected keyword argument 'rounding_mode'` @yhna940

```python
import torch
import math
import onnx
import onnxruntime
from onnxruntime.quantization import QuantizationMode

def pixel_unshuffle(x, scale):
    """Down-sample by pixel unshuffle.

    Args:
        x (Tensor): Input tensor.
        scale (int): Scale factor.

    Returns:
        Tensor: Output tensor.
    """

    b, c, h, w = x.shape
    h = h // scale
    w = w // scale
    x = x.view([b, c, h, scale, w, scale])
    x = x.permute(0, 1, 3, 5, 2, 4)
    return x.reshape(b, -1, h, w)

class PixelUnshuffleModel(torch.nn.Module):
    def __init__(self, scale):
        super().__init__()
        self.scale = scale

    def forward(self, x):
        y = pixel_unshuffle(x, self.scale)
        return y

def create_pixel_unshuffle_model(scale):
    model = PixelUnshuffleModel(scale)
    dummy_input = torch.randn(1, 3, 256, 256)
    inputs = (dummy_input,)
    output = model(*inputs)
    input_names = ['input']
    output_names = ['output']
    dynamic_axes = {'input': {0: 'batch_size', 2: 'height', 3: 'width'}, 'output': {0: 'batch_size', 2: 'height', 3: 'width'}}
    return model, inputs, output, input_names, output_names, dynamic_axes


model, inputs, output, input_names, output_names, dynamic_axes = create_pixel_unshuffle_model(scale=2)
torch.onnx.export(model, inputs, 'pixel_unshuffle.onnx', input_names=input_names, output_names=output_names,
                  dynamic_axes=dynamic_axes, opset_version=11)


sess = onnxruntime.InferenceSession('pixel_unshuffle.onnx')
x = torch.randn(1, 3, 128, 128)
input_name = sess.get_inputs()[0].name
ort_output = sess.run(None, {input_name: x.numpy()})

torch_output = model(x)
print(torch.allclose(torch_output, torch.tensor(ort_output[0]), rtol=1e-03, atol=1e-05))
```

Thanks for correcting my mistake. The problem is that the lowest version you support, Torch (1.6), does not support rounding mode for division operations. So I changed it to use a simpler floor div to track the shape of the tensor (since the assert statement guarantees that it is divisible by scale)! thank you. The following script applied the onnx export for dynamic input to the modified model and it worked. thank you.

Z-Fran (approved):

PR #1645, opened 2023-02-15
[Fix] Enhance the compatibility of training stylegan 2

HAOCHENYE:
## Motivation

PyTorch consolidated convolution backward API in [PR](https://github.com/pytorch/pytorch/commit/3dc3651e0ee3623f669c3a2c096408dbc476d122). It means users cannot train styegan2 with PyTorch version higher than 1.10.1

This PR enhances the compatibility of the code referring to the [discussion](https://github.com/pytorch/pytorch/issues/74437)

Resolve [#3](https://github.com/open-mmlab/mmgeneration/issues/322)

## Modification

Please briefly describe what modification is made in this PR.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

- [ ] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmediting/blob/master/.github/CONTRIBUTING.md) to create this PR.
- [ ] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmediting/blob/master/.github/CONTRIBUTING.md) are used to fix the potential lint issues.
- [ ] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [ ] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [ ] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [ ] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects.
- [ ] CLA has been signed and all committers have signed the CLA in this PR.

### DIFF
--- a/configs/styleganv2/stylegan2_c2_8xb4_ffhq-1024x1024.py
+++ b/configs/styleganv2/stylegan2_c2_8xb4_ffhq-1024x1024.py
@@ -42,7 +42,7 @@ optim_wrapper = dict(
                                                         0.99**d_reg_ratio))))
 
 batch_size = 4
-data_root = 's3://openmmlab/datasets/editing/ffhq/images'
+data_root = './data/ffhq/images'
 
 train_dataloader = dict(
     batch_size=batch_size, dataset=dict(data_root=data_root))
--- a/configs/styleganv2/stylegan2_c2_8xb4_ffhq-1024x1024.py
+++ b/configs/styleganv2/stylegan2_c2_8xb4_ffhq-1024x1024.py
@@ -42,7 +42,7 @@ optim_wrapper = dict(
                                                         0.99**d_reg_ratio))))
 
 batch_size = 4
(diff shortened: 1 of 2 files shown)


configs/styleganv2/stylegan2_c2_8xb4_ffhq-1024x1024.py
```diff
@@ -42,7 +42,7 @@
                                                         0.99**d_reg_ratio))))
 
 batch_size = 4
-data_root = './data/ffhq/images'
+data_root = 's3://openmmlab/datasets/editing/ffhq/images'
```
LeoXing1996, line 45: Please keep this config unchanged since this path only works in the internal network. 
Noted that, you can generate configs with *ceph path* by `.dev_scripts/create_ceph_configs.py`

PR #1686, opened 2023-03-04
Make the image order in VisualizationHook as it is in the config

zeakey:
## Motivation

Save visual images according to the order in the config.

Currently the images are concatenated according to the order in the results dict, and the order is not configurable:

<https://github.com/open-mmlab/mmediting/blob/d18fc44868082b3bff4d80973b09b1f2f2db88a3/mmedit/core/hooks/visualization.py#L75>

I changed this line to 

```
img_list = [results[k] for k in self.res_name_list if k in results]
```
so that we can configure the order of images being saved using  `visual_config.res_name_list`.

### DIFF
--- a/mmedit/core/hooks/visualization.py
+++ b/mmedit/core/hooks/visualization.py
@@ -72,7 +72,7 @@ class MMEditVisualizationHook(Hook):
 
         filename = self.filename_tmpl.format(runner.iter + 1)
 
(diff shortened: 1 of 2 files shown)


LeoXing1996 (approved):
@zeakey This is a good feature. Thanks for your contribution!

PR #1899, opened 2023-06-10
[Fix] Requirements

ryanxingql:
## Remove `mmcv` and `mmengine` from "requirements/runtime.txt"

`mmcv` and `mmengine` are repeatedly listed in "requirements/{mminstall.txt,runtime.txt}". Following [MMDetection](https://github.com/open-mmlab/mmdetection/tree/main/requirements), we removed `mmcv` and `mmengine` from "requirements/runtime.txt", and kept those in "requirements/mminstall.txt".

The reason is that we have manually installed MMCV and MMEngine before installing MMagic as indicated in the [Doc](https://github.com/open-mmlab/mmagic/blob/main/docs/en/get_started/install.md). As a result, there is no need to require `mmcv` and `mmengine` in "requirements/runtime.txt", which are to be installed with "setup.py".

After this modification, the installation of MMagic is much faster.

## Add dependencies Click and pandas in "requirements/runtime.txt"

We added `Click` and `pandas` in "requirements/runtime.txt". They are required by "mmagic/utils/io_utils.py" and "mmagic/models/editors/disco_diffusion/guider.py".

These two packages are installed when installing [MIM](https://github.com/open-mmlab/mim/blob/main/requirements/install.txt). However, when someone uses pip instead of MIM to install MMCV, these two packages are missed and cannot be found.

### DIFF
--- a/requirements/mminstall.txt
+++ b/requirements/mminstall.txt
@@ -1,2 +1,2 @@
-mmcv>=2.0.0
+mmcv>=2.0.0rc1
 mmengine>=0.4.0
--- a/requirements/mminstall.txt
+++ b/requirements/mminstall.txt
@@ -1,2 +1,2 @@
-mmcv>=2.0.0rc1
+mmcv>=2.0.0
 mmengine>=0.4.0
--- a/mmagic/__init__.py
+++ b/mmagic/__init__.py
@@ -20,7 +20,7 @@ except ImportError:
         return digit_ver
 
 
-MMCV_MIN = '2.0.0rc1'
+MMCV_MIN = '2.0.0'
 MMCV_MAX = '2.1.0'
 mmcv_min_version = digit_version(MMCV_MIN)
 mmcv_max_version = digit_version(MMCV_MAX)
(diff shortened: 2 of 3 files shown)


requirements/mminstall.txt
```diff
@@ -1,2 +1,2 @@
-mmcv>=2.0.0
+mmcv>=2.0.0rc1
```
Z-Fran, line 1: Please use 'mmcv>=2.0.0'

Z-Fran (dismissed):

ryanxingql:
@Z-Fran Hi, I notice that in "mmagic/mmagic/engine/__init__.py" the minimal version of MMCV is asserted to be "2.0.0rc1".

Z-Fran:
> @Z-Fran Hi, I notice that in [mmagic/mmagic/engine/**init**.py](https://github.com/open-mmlab/mmagic/blob/main/mmagic/__init__.py), the minimal version of MMCV is asserted to be "2.0.0rc1".

Actually, it should also be modified to 2.0.0 here.

ryanxingql:
> > @Z-Fran Hi, I notice that in [mmagic/mmagic/engine/**init**.py](https://github.com/open-mmlab/mmagic/blob/main/mmagic/__init__.py), the minimal version of MMCV is asserted to be "2.0.0rc1".
> 
> Actually, it should also be modified to 2.0.0 here.

May I change all these "2.0.0rc1" to "2.0.0"?

Z-Fran (approved):

Issue #1927, opened 2023-07-10
[Bug] `num_images_per_prompt` argument does not work in Controlnet


gvalvano:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmagic/issues) and [Discussions](https://github.com/open-mmlab/mmagic/discussions) but cannot get the expected help.
- [X] I have read the [FAQ documentation](https://mmagic.readthedocs.io/en/latest/faq.html) but cannot get the expected help.
- [X] The bug has not been fixed in the [latest version (main)](https://github.com/open-mmlab/mmagic) or [latest version (0.x)](https://github.com/open-mmlab/mmagic/tree/0.x).

### Task

I'm using the official example scripts/configs for the officially supported tasks/models/datasets.

### Branch

main branch https://github.com/open-mmlab/mmagic

### Environment

The argument  `num_images_per_prompt` of ControlNet's infer method does not work with values different from the default ([line](https://github.com/open-mmlab/mmagic/blob/1c932c024ccb88347618abfdfa61756f49646002/mmagic/models/editors/controlnet/controlnet.py#L409) of code). Instead, it works as expected if we use the default value: `num_images_per_prompt=1`.

### Reproduces the problem - code sample

```python

# checkpoint_file and config for the model:
root_dir = "path/to/root/dir"
config_file = f"{root_dir}/configs/controlnet.py"
checkpoint_file = f"{root_dir}/logs/controlnet/iter_10000.pth"

controlnet = init_model(config_file, checkpoint_file, device='cuda:0')

# get the input data for the model:
ctrl= mmcv.imread(filename, channel_order='rgb')
prompt = "input prompt"

prediction_dict = model.infer(
    prompt=prompt, 
    control=Image.fromarray(ctrl), 
    controlnet_conditioning_scale=1.0,
    num_inference_steps=20,
    guidance_scale=3,
    negative_prompt="low quality, saturated image, fake",
    num_images_per_prompt=2  # <----------------------------------- This raises the error when we use != 1
)     
```

### Reproduces the problem - command or script

See code snipped above.

### Reproduces the problem - error message

```error
│ /opt/conda/lib/python3.8/site-packages/torch/autograd/grad_mode.py:27 in decorate_context        │
│                                                                                                  │
│    24 │   │   @functools.wraps(func)                                                             │
│    25 │   │   def decorate_context(*args, **kwargs):                                             │
│    26 │   │   │   with self.clone():                                                             │
[...]

PR #1929, opened 2023-07-11
[Fix] 'num_images_per_prompt' argument does not work in Controlnet

yayoyo66:
## Motivation

fix the bug [link](https://github.com/open-mmlab/mmagic/issues/1927)

## Modification

modify the function **prepare_control** in  'mmagic/models/editors/controlnet/controlnet.py'

## BC-breaking (Optional)

I guess no


LeoXing1996:
Hey @yayoyo66, thanks for your controbution. We provide a better and more elegant fix in PR #1936. 
Therefore we close this PR.

PR #2008, opened 2023-08-29
[CodeCamp2023-644] New config for cyclegan

jianyuhaohai:
When I'm working on this task, I have a few questions:

    In base_cyclegan.py, I couldn't locate the position of the normal class for init_cfg=dict(type='normal', gain=0.02)).

    When running train.py, whether with the old config or the new config, I encounter an UnboundLocalError: local variable 'init_info' referenced before assignment.


liuwenran:
> When I'm working on this task, I have a few questions:
> 
> ```
> In base_cyclegan.py, I couldn't locate the position of the normal class for init_cfg=dict(type='normal', gain=0.02)).
> 
> When running train.py, whether with the old config or the new config, I encounter an UnboundLocalError: local variable 'init_info' referenced before assignment.
> ```

init_cfg=dict(type='normal') is the parameter for class PatchDiscriminator.
Please see 'mmagic/mmagic/models/archs/patch_disc.py'

When running train.py, old config and new config should both work.

PR #2030, opened 2023-09-14
[CodeCamp2023-653] Add new configs of Real BasicVSR

RangeKing:
## Motivation

Add new configs of Real BasicVSR

## Result

- The output of `config_validate.py`
config_validate_vsr

- The output of `test.py` (only use the folder `000` of VideoLQ for test)
config_validate_vsr_2


LeoXing1996 (approved):
LGTM!

PR #2042, opened 2023-09-22
[CodeCamp2023-645]Add dreambooth new cfg

YanxingLiu:
Add new cfg for dreambooth. 
All new confighave been verified by the validation scripts.
Generated fake images:
fake_img_1000


liuwenran (approved):
Thank you for your contribution!

